PR #4, opened 2019-02-27
Add create-tables and improve restore process

anuriq:
- add create-tables command implementation

when restoring to fresh installation of Clickhouse a step to recreate databases and tables had to be done manually. now create-tables method takes metadata information stored with backup and recreates all objects including distributed tables (they are created last). tables should not exist before running this command.

- change convert partition method to support custom partitioning key

convertPartition was expecting one of two:
    - legacy partitioning key toYYYYMM(date_column)
    - partitioning key based on data YYYY-MM-DD

as cited in documentation https://clickhouse.yandex/docs/en/operations/table_engines/custom_partitioning_key/ name of partition same as in system.parts table can be used. this is also same as used in file name. this means just that part itself may be used in ATTACH PARTITION query.

- make partition size for S3 object configurable

there was a hard coded value equal to 5 MB. this is saved as default value, but now it may be changed via configuration file.

- add documentation comments and fix some comments. no code changes


anuriq:
going to add code and recreate PR

PR #7, opened 2019-03-06
Some improvements for corner cases

anuriq:
## do not allow freeze if shadow dir is not empty

Current behaviour leads to some sophisticated cases during restore: if
shadow dir already had any partitions, those are either old, because
partitions were overwritten, or same and a new hard link will be
created. In both of these cases files will be uploaded to S3. During
restore there is no method to distinguish such parts.
So, we need to have shadow dir empty, to have consistent current copy
of data in Clickhouse. This can be easily done via clean command.

## do not throw error if there are no partitions in Clickhouse

In case there are no actual data (partitions) in Clickhouse just log
a message about this and quit.

## do not throw error during restore if there are no partitions

In case a backup was made which does not have partitions (shadow dir is
empty) but metadata is there and may be used for creating tables, we
just log a message, that there is nothing to do.

## do not throw error during clean if shadow dir is absent and move cleanDir function to utils

### DIFF
--- a/main.go
+++ b/main.go
@@ -214,16 +214,9 @@ func createTables(config Config, args []string, dryRun bool) error {
 	}
 	defer ch.Close()
 
-	dataPath := config.ClickHouse.DataPath
-	if dataPath == "" {
-		if err := ch.Connect(); err != nil {
-			return fmt.Errorf("can't connect to clickhouse for get data path with: %v\nyou can set clickhouse.data_path in config", err)
-		}
-		defer ch.Close()
-		var err error
-		if dataPath, err = ch.GetDataPath(); err != nil || dataPath == "" {
-			return fmt.Errorf("can't get data path from clickhouse with: %v\nyou can set data_path in config file", err)
-		}
+	dataPath, err := ch.GetDataPath()
+	if err != nil || dataPath == "" {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -214,16 +214,9 @@ func createTables(config Config, args []string, dryRun bool) error {
...
-			return fmt.Errorf("can't connect to clickhouse for get data path with: %v\nyou can set clickhouse.data_path in config", err)
-		}
-		defer ch.Close()
-		var err error
-		if dataPath, err = ch.GetDataPath(); err != nil || dataPath == "" {
-			return fmt.Errorf("can't get data path from clickhouse with: %v\nyou can set data_path in config file", err)
-		}
+	dataPath, err := ch.GetDataPath()
+	if err != nil || dataPath == "" {
+		return fmt.Errorf("can't get data path from clickhouse with: %v\nyou can set data_path in config file", err)
```
AlexAkulov, line 219: Если data_path установлен в конфиге он должен иметь больший приоритет чем тот который возвращает КХ. 
Например, это нужно, если КХ запущен в докере, а бэкап запускается с хоста. 
Кажется эти изменения приводят к тому, что data_path из конфига вообще не читаются.

anuriq: да, но в функции GetDataPath первое что вернется это как раз значение из конфига же

AlexAkulov: ok

PR #50, opened 2019-10-09
Support for SSE AES256 on S3

pelletier:
This pull request adds support for [server-side encryption](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html) on S3 buckets.

Note: those changes only allow for AES256 encryption. When using aws:kms, additional arguments (keys) will need to be passed. This is outside the scope of this PR.


AlexAkulov:
Thank you!

PR #60, opened 2019-11-13
install from sources doesn't work — fix command

develar:
`GO111MODULE=on` must be used, otherwise go is not able to build correctly.

You are not alone — see https://github.com/google/ko#installation for example.


AlexAkulov:
Thanks

PR #64, opened 2019-11-21
Added the option to skip certificate validation

martenlindblad:
Added the option to skip certificate validation when using S3. We have a few internal S3 applicances that don't have valid certificates. This pull requests makes it possible for us to backup to those S3 appliances.


AlexAkulov:
Thank you!

PR #77, opened 2020-02-12
Return go sum

### LINKED ISSUE #76: Why go.sum is ignored?
Hello, and thank you for the great software.

I have a question regarding https://github.com/AlexAkulov/clickhouse-backup/commit/68dfb06c37b4b2ea31948e841ae79f4f06078a31

According to https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file, this file should be in the repository, to guarantee the reproducible build and avoid suspicious sources changes.

I'd like to create a PR to return it back if you don't mind.

Felixoid:
Following the issue, fixes #76 

Sorry for an issue, it would be easier to just create a PR


AlexAkulov:
Thank you!

PR #97, opened 2020-06-10
Separated clickhouse data and backup path

DrissiReda:
In some usecases (namely kubernetes for myself). You might want to avoid writing to `/var/lib/clickhouse` externally, so I think that separating data and backup path is useful. 

In my case I'm limiting the clickhouse-backup docker to only have read permissions on /var/lib/clickhouse so I can't add backups there, nor can I clean the "shadow" folder. Is leaving the shadow folder and overwriting it every time I create a backup a problem?

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -101,6 +101,7 @@ clickhouse:
   port: 9000                   # CLICKHOUSE_PORT
   timeout: 5m                  # CLICKHOUSE_TIMEOUT
   data_path: ""                # CLICKHOUSE_DATA_PATH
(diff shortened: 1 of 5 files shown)


AlexAkulov:
Sorry, this PR is outdated.

DrissiReda:
Do you need me to update it? Or is this feature irrelevant now.

AlexAkulov:
I think this feature is more difficult thaN it seems at first look. You must have option to define backup path for THE ENTIRE disk in multidisk installations of ClickHouse and you have to make sure that backup paths and clickhouse data ARE located on the same mount points.

PR #112, opened 2020-08-23
Add Azure block blobs remote storage backend

atykhyy:
This PR adds Azure block blobs as remote storage backend for backups. Server-side encryption is supported with the help of Azure/azure-storage-blob-go/tree/feature/clientprovidedkey feature branch. I had to vendor two files from this branch (pending Azure/azure-storage-blob-go/pull/202) because their UploadStreamToBlockBlob method omits CPK support.


AlexAkulov:
Thank you!

PR #114, opened 2020-09-02
azblob: don't require container creation rights

atykhyy:


AlexAkulov:
Не совсем понял что там за рандомный test blob name создаётся и зачем?

atykhyy:
That's to test whether the tool has access rights to the container. It may not have enough rights to _create_ a container, and it doesn't really need to create containers, just read and write blobs in the given container. This way the tool can be given a shared access signature with rights to read and write blobs in a specific container, which is good from the POV of least privilege. I don't think it's possible to create a SAS which allows one to create a specific container and not create/delete any containers in the storage account.

AlexAkulov:
Thank you! 
I added this changes in code.

PR #168, opened 2021-03-16
when downloading backups, avoid creating local backup dir before conf…

ElPeque:
…irming it exists remotely

This is causing trying to download non-existing backups to create empty/fake local backups. So any typo in the name or any error will create an invalid backup entry.

It will still be created below as part of: `os.MkdirAll(extractDir, os.ModePerm)`


AlexAkulov:
Thank you!

PR #192, opened 2021-05-04
fixes for restore corner cases

Slach:
- add integration test for MATERIALIZED VIEW + .inner table
- change sort order during `restore --rm`, I hope it allow to resolve depency object restore


pkg/clickhouse/clickhouse.go
```diff
@@ -488,11 +503,16 @@ func (ch *ClickHouse) CreateDatabase(database string) error {
...
+	var isAtomic bool
+	var err error
+	if isAtomic, err = ch.IsAtomic(table.Database); err != nil {
 		return err
 	}
 	if dropTable {
-		query := fmt.Sprintf("DROP TABLE IF EXISTS `%s`.`%s` NO DELAY", table.Database, table.Name)
+		query := fmt.Sprintf("DROP TABLE IF EXISTS `%s`.`%s`", table.Database, table.Name)
+		if isAtomic {
+			query += " SYNC"
```
Slach: NO DELAY here

PR #200, opened 2021-05-27
add backup_name/metadata.json, databases section which store CREATE DATABASE sql queries for properly restore

Slach:
this PR fix corner cases, when we have settings `default_database_engine`, restore databases as "CREATE DATABASE IF NOT EXISTS db" works wrong, they will restore Ordinary as Atomic, and Atomic as Ordinary

old backups which created with old version will restore with old behavior


pkg/backup/restore.go
```diff
@@ -45,6 +45,11 @@ func Restore(cfg *config.Config, backupName string, tablePattern string, schemaO
 		if err := json.Unmarshal(backupMetadataBody, &backupMetadata); err != nil {
 			return err
 		}
+		for _, database := range backupMetadata.Databases {
+			if err := ch.CreateDatabaseFromQuery(database.Query); err != nil {
```
Slach, line 49: there is main part for fix

PR #258, opened 2021-09-03
Fix restore command failure

obazna:
I used s3 for data storage and faced with `restore` command failure.  When I tried to restore my backup I received the next error  ```error can't attach partitions for table 'default.s3_table': code: 233, message: Detached part "all_1_1_0" not found```

Server's config related to data storage:
```yaml
storage_configuration:
    # S3 Credentials
    disks:
        s3:
            type: s3
            endpoint: https://my-end-point.s3.us-east-1.amazonaws.com/   
    policies:
        s3only:
            volumes:
                main:
                    disk: s3
```

The thing is that by default s3 disk will have the next paths:
**metadata_path** - ```/var/lib/clickhouse/disks/s3/```
**cache_path** - ```/var/lib/clickhouse/disks/s3/cache/```
etc.

As you see above you will have two disks:
```'default'``` - with path - ```/var/lib/clickhouse/```
```'s3'``` - with path - ```/var/lib/clickhouse/disks/s3/```

And the thing is that ```s3``` path is inside ```default``` disk path. And unfortunately **GetDiskByPath** returns the first disk name with a prefix that matches to the input path. In my case it's ```default``` instead of ```s3```.  And then everything go wrong. 
So basically I fixed that error and now the function correctly determines the disk.

### DIFF
--- a/pkg/clickhouse/utils.go
+++ b/pkg/clickhouse/utils.go
@@ -9,12 +9,26 @@ import (
 )
 
 func GetDiskByPath(disks []Disk, dataPath string) string {
+	prefixLengthMap := map[int]string{}
+
 	for _, disk := range disks {
 		if strings.HasPrefix(dataPath, disk.Path) {
-			return disk.Name
+			prefixLengthMap[len(disk.Path)] = disk.Name
 		}
 	}
-	return "unknown"
+
+	if len(prefixLengthMap) == 0 {
+		return "unknown"
+	}
+
+	var maxPrefixLength int
+	for prefixLength := range prefixLengthMap {
+		if prefixLength > maxPrefixLength {
+			maxPrefixLength = prefixLength
+		}
+	}
+
+	return prefixLengthMap[maxPrefixLength]
 }
 
 func GetDisksByPaths(disks []Disk, dataPaths []string) map[string]string {

### CODE REVIEW

**Review by Slach — APPROVED** [review]
looks good
did you test how exactly works ALTER TABLE ... FREEZE on data parts which stored on S3 disk?
is it just create a hard link with parts metadata?

File: pkg/clickhouse/utils.go [hunk 1]
```diff
@@ -9,12 +9,26 @@ import (
 )
 
 func GetDiskByPath(disks []Disk, dataPath string) string {
+	prefixLengthMap := map[int]string{}
+
 	for _, disk := range disks {
 		if strings.HasPrefix(dataPath, disk.Path) {
-			return disk.Name
+			prefixLengthMap[len(disk.Path)] = disk.Name
```
AlexAkulov, line 16: When we have multiple disks with following map: disk1=/var/lib/clickhouse/disk1, disk2=/var/lib/clickhouse/disk2, disk3=/var/lib/clickhouse/disk3 then I'm not sure we will restore data properly. 
Could you add integration tests which cover this corner case?

AlexAkulov (changes requested):
When we have multiple disks with following map: disk1=/var/lib/clickhouse/disk1, disk2=/var/lib/clickhouse/disk2, disk3=/var/lib/clickhouse/disk3 then I'm not sure we will restore data properly.
Could you add integration tests which cover this corner case?

pkg/clickhouse/utils.go
obazna: I did a lot of manual testing. It restores everything perfectly. But I've tested with s3 as the main storage only. The tool has to restore paths like you mentioned because the default clickhouse paths looks like:
```metadata_path — Path on local FS to store metadata files for S3. Default value is /var/lib/clickhouse/disks/<disk_name>/.```

```data - var/lib/clickhouse/disks/<disk_name>/data```

See for[ more details.](https://clickhouse.tech/docs/en/engines/table-engines/mergetree-family/mergetree/)

I'll come up with intergration tests when I have time. But feel free to help if it's ASAP.

Slach:
@obazna could you check `alexakulov/clickhouse-backup:master` docker image with your configuration?

PR #263, opened 2021-09-09
RequiredBackup JSON field in /backup/list API call

vanyasvl:
RequiredBackup JSON field added in /backup/list API call to match CLI "list" command output


Slach:
thanks a lot for contributing

Issue #197, opened 2021-05-19
azblob should support system assigned identity


roman-vynar:
Hello,

First of all, Thanks for the tool!

There are 4 methods to auth to get into az storage account:
1. account key
2. SAS
3. oauth credentials (AD > App registrations > credentials: app id and secret token)
4. system assigned identity

clickhouse-backup supports only 1-2 methods.

4th method is when you assign a role (e.g. Storage Blob Data Contributor) to the Service Principal of a VM instance on a scope of some blob container. Then from that instance you can do "az login --identity" (in case of az CLI) and just work directly with the container w/o any credentials. This is the simplest method and more secure than sharing an account key for example (1st method).

Slach:
closed by https://github.com/AlexAkulov/clickhouse-backup/pull/267

PR #267, opened 2021-09-15
Support of ManagedIdentity azure auth

roman-vynar:
First of all, thanks for such a great tool!

We are using this tool with both AWS S3 and Azure blob storage.

AWS S3 works ok with IAM Instance Role permissions associated with an instance, w/o using any static keys.

However, with Azure the tool only works with Storage Account access keys (not really secure method, providing full access to SA, the keys are static and so on) or SAS tokens (which are not flexible to use and insecure the way Azure implemented them).

There are 2 other methods to access Azure blob storage: OAuth credentials of AD app registrations (tenant id + app id + client secret token) or using ManagedIdentity. 

The last one is nice because it can work with SystemAssigned Identity associated with a VM, much like Instance Role with AWS. You can grant an access to a specific container within Storage Account to a service principal of a single VM without using any kind of credentials or static keys.

Adding this simple PR to be able to use ManagedIdentity. Just grant access and set "use_managed_identity: true".

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -126,6 +126,7 @@ azblob:
   account_name: ""             # AZBLOB_ACCOUNT_NAME
   account_key: ""              # AZBLOB_ACCOUNT_KEY
   sas: ""                      # AZBLOB_SAS
+  use_managed_identity: false  # AZBLOB_USE_MANAGED_IDENTITY
   container: ""                # AZBLOB_CONTAINER
   path: ""                     # AZBLOB_PATH
   compression_level: 1         # AZBLOB_COMPRESSION_LEVEL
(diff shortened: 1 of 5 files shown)


rhuddleston:
Fixes https://github.com/AlexAkulov/clickhouse-backup/issues/197

Slach:
@rhuddleston great contribution thanks a lot

How did you test it?
Did you try to change azure config on `./tests/intergations/config-azure.yaml` 
and run `AZURE_TESTS=1 bash -xe ./tests/intergations/run.sh`?

roman-vynar:
Hello, I have tested this straight on our environment, on Azure VM and it works ok.

My config:
```
azblob:
  endpoint_suffix:      core.windows.net
  account_name:         foobar
  account_key:          ""
  sas:                  ""
  use_managed_identity: true
  container:            backups-clickhouse
  path:                 "clickhouse/binary"
  compression_level:    1
  compression_format:   gzip
  sse_key:              ""
```

Here is how do I grant the access for dbnode1 to "backups-clickhouse" container in "foobar" storage account using Terraform Azure plugin:
```
data "azurerm_storage_container" "backups_cont" {
  name                 = "backups-clickhouse"
  storage_account_name = "foobar"
}

resource "azurerm_role_assignment" "dbnode_backup_access" {
  scope                = data.azurerm_storage_container.backups_cont.resource_manager_id
  role_definition_name = "Storage Blob Data Owner"
  principal_id         = azurerm_linux_virtual_machine.dbnode1.identity[0].principal_id
}
```

roman-vynar:
Thank you!

Issue #273, opened 2021-09-23
Make S3 partSize as config option


vanyasvl:
Default partSize calculation `partSize := cfg.General.MaxFileSize / 10000` isn't work good with some S3 compatible storages like Swift. Swift has a low performance with a small chunks and doesn't accept chunks more than 5Gb. 

So it will be good to have S3 partSize in config.

Slach:
thanks a lot for reporting, it make sense, and could be implements

PR #276, opened 2021-09-30
S3_PART_SIZE config option

alschastny:
**REALTED**
Issue #273

**PROBLEM**
I'am using MinIO as S3 object storage for ClickHouse backups, and I can't upload big parts to it, because of it "Maximum number of parts per upload" limit (https://docs.min.io/docs/minio-server-limits-per-tenant.html). I've got the following lines from log file:

**LOG**
`-----------------------------------------------------
2021/09/30 06:20:02  info DEBUG: Request s3/UploadPart Details:
---[ REQUEST POST-SIGN ]-----------------------------
PUT /PATH?partNumber=10000&uploadId=d0a6d3b1-e2ce-4a12-912c-65ad69727ba9 HTTP/1.1
Host: 100.100.100.100:9000
User-Agent: aws-sdk-go/1.40.31 (go1.16.7; linux; amd64) S3Manager
Content-Length: 10737418
Authorization: AWS4-HMAC-SHA256 Credential=clickhouse_backup/20210930/us-east-1/s3/aws4_request, SignedHeaders=content-length;content-md5;host;x-amz-content-sha256;x-amz-date, Signature=cc81e19deed144659ae2cb5dd6b9acadb8516aae5a331f58e644e4ea5c8e7aa2
Content-Md5: wouhtJAHBQ+phEjQYmqVng==
Expect: 100-Continue
X-Amz-Content-Sha256: c13d660237fb3f3362c26888da1e26fb2208f266d33d0b0dd76c0d75067f3513
X-Amz-Date: 20210930T062002Z
Accept-Encoding: gzip

-----------------------------------------------------
2021/09/30 06:20:02  info DEBUG: Response s3/UploadPart Details:
---[ RESPONSE ]--------------------------------------
HTTP/1.1 200 OK
Content-Length: 0
Accept-Ranges: bytes
Content-Security-Policy: block-all-mixed-content
Date: Thu, 30 Sep 2021 06:20:02 GMT
Etag: "c28ba1b49007050fa98448d0626a959e"
Server: MinIO
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Origin
Vary: Accept-Encoding
X-Amz-Request-Id: 16A9853508EA0DD9
X-Content-Type-Options: nosniff
X-Xss-Protection: 1; mode=block

-----------------------------------------------------
2021/09/30 06:20:02  info DEBUG: Request s3/AbortMultipartUpload Details:
---[ REQUEST POST-SIGN ]-----------------------------
DELETE /PATH/default_2.tar?uploadId=d0a6d3b1-e2ce-4a12-912c-65ad69727ba9 HTTP/1.1
Host: 100.100.100.100:9000
User-Agent: aws-sdk-go/1.40.31 (go1.16.7; linux; amd64) S3Manager
Authorization: AWS4-HMAC-SHA256 Credential=clickhouse_backup/20210930/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=24a505afe1bc5f3cac70dbbd7c96e7b4eb77a243e20e152401ab9f90aff77b0b
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20210930T062002Z
Accept-Encoding: gzip

-----------------------------------------------------
2021/09/30 06:20:12  info DEBUG: Response s3/AbortMultipartUpload Details:
---[ RESPONSE ]--------------------------------------
HTTP/1.1 204 No Content
Accept-Ranges: bytes
Content-Security-Policy: block-all-mixed-content
Date: Thu, 30 Sep 2021 06:20:12 GMT
Server: MinIO
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Origin
Vary: Accept-Encoding
X-Amz-Request-Id: 16A985350E766F62
X-Content-Type-Options: nosniff
X-Xss-Protection: 1; mode=block

-----------------------------------------------------`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -82,6 +82,7 @@ type S3Config struct {
 	DisableCertVerification bool   `yaml:"disable_cert_verification" envconfig:"S3_DISABLE_CERT_VERIFICATION"`
 	StorageClass            string `yaml:"storage_class" envconfig:"S3_STORAGE_CLASS"`
 	Concurrency             int    `yaml:"concurrency" envconfig:"S3_CONCURRENCY"`
+	PartSize                int64  `yaml:"part_size" envconfig:"S3_PART_SIZE"`
 	Debug                   bool   `yaml:"debug" envconfig:"S3_DEBUG"`
 }
 
@@ -316,6 +317,7 @@ func DefaultConfig() *Config {
 			DisableCertVerification: false,
 			StorageClass:            s3.StorageClassStandard,
 			Concurrency:             1,
+			PartSize:                0,
 		},
 		GCS: GCSConfig{
 			CompressionLevel:  1,
--- a/pkg/new_storage/general.go
+++ b/pkg/new_storage/general.go
@@ -403,9 +403,12 @@ func NewBackupDestination(cfg *config.Config) (*BackupDestination, error) {
 			cfg.General.DisableProgressBar,
 		}, nil
 	case "s3":
-		partSize := cfg.General.MaxFileSize / 10000
-		if partSize < 5242880 {
-			partSize = 5242880
+		partSize := cfg.S3.PartSize
+		if cfg.S3.PartSize <= 0 {
+			partSize = cfg.General.MaxFileSize / 10000
+			if partSize < 5242880 {
+				partSize = 5242880
+			}
 		}
 		s3Storage := &S3{
 			Config:      &cfg.S3,
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -149,6 +149,8 @@ s3:
   sse: AES256                      # S3_SSE
   disable_cert_verification: false # S3_DISABLE_CERT_VERIFICATION
   storage_class: STANDARD          # S3_STORAGE_CLASS
+  # if less or eq 0 then calculated as max_file_size / 10000 
+  part_size: 0                     # S3_PART_SIZE
   debug: false                     # S3_DEBUG
 gcs:
   credentials_file: ""         # GCS_CREDENTIALS_FILE


Slach (approved):

pkg/new_storage/general.go
```diff
@@ -403,15 +403,11 @@ func NewBackupDestination(cfg *config.Config) (*BackupDestination, error) {
 			cfg.General.DisableProgressBar,
 		}, nil
 	case "s3":
-		partSize := cfg.General.MaxFileSize / 10000
```
Slach, line 406: These changes limit maximum upload size to 50Gb for original S3
let's adjust your PR

* setup default configuration `PartSize` to zero it mean no restriction to part size
* return code which calculate partSize 
* add 
```go
if cfg.S3.PartSize > 0 && partSize > cfg.S3.PartSize { 
  partSize = cfg.S3.PartSize
}
```

config/config.go
```diff
@@ -316,6 +317,7 @@ func DefaultConfig() *Config {
 			DisableCertVerification: false,
 			StorageClass:            s3.StorageClassStandard,
 			Concurrency:             1,
+			PartSize:                5 * 1024 * 1024, // 5MB
```
Slach: lets setup it zero  by default

Slach (changes requested):
Thanks a lot for your contributing

Could you adjust this PR a little bit?

Do you have configured limits on your MINIO?
According to your link
`minio` allow by default the same part size limits as original S3
it means the `minio` allow `partSize` between 5Mb and 5Gb

in your logs you try to upload only 10737418 bytes, it not so much

pkg/new_storage/general.go
Slach: Maybe we also should add max limits to 5Gb for calculated `partSize`

alschastny: https://github.com/AlexAkulov/clickhouse-backup/blob/0bc5d7c5aaef304aea09d429a7b853940cf0c825/pkg/backup/upload.go#L242
cfg.General.MaxFileSize config parameter limits the "preferred" size of uploaded files, but shouldn't limit the upload process at all, as I see it.

So, if we have files in our backup that are larger than MaxFileSize, should we abort the backup process altogether? I think no. I don't want to increase MaxFileSize to high values because that will create big parts to upload.
If we will have 2 independent settings for the preferred size of the uploaded file and for setting the size of the chunk, we will get a more flexible solution - we will be able to set both the preferred size of the uploaded file and the maximum size of the file uploaded to S3. I agree to set PartSize to 0 by default and calculate the default value in this case.

I've already updated my PR to described logic.

Slach: `cfg.General.MaxFileSize` have not clear mental name,
it limits archive size on remote storage, and we split data parts by this parameter, to avoid too big archive sizes
if we have one part more than `cfg.General.MaxFileSize` it still tries to upload, but just in the separate `.tar` file on remote storage

Slach: > I don't want to increase MaxFileSize to high values because that will create big parts to upload.

You don't need to increase the `MaxFileSize` which related to max size of archive on remote storage (one table -> one and more archive which depends on disks mapping and table size)

you should increase `partSize`, big `partSize` for multipart S3 upload increase upload speed on S3, 
if you agree  with my proposal, you can limit max `partSize` for data part during s3 multipart upload to `Minio`

I ask again, did you change anything on `minio` side? By default, it allows the same limits for multipart upload as original S3.

In real cases, if you have `0.5TB+` table you can have 50Gb+ data parts in one table

Slach (approved):

pkg/new_storage/general.go
alschastny: No, I I haven't changed anything in MinIO settings.
As we can see from the [S3 dcos](https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html) they have the same 10,000 parts upload limit. If we know this limitation for S3, it would be nice to display an error and not try to upload files for which: **file_size/part_size>10000**.
At the moment, you can only find out about this by enabling the S3_DEBUG setting.

PR #328, opened 2021-12-16
Allow using OIDC token for AWS credentials

ldx:
EKS on AWS uses OIDC with web identity tokens to allow pods to assume an IAM role via their service accounts: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html. Pods have two environment variables automatically exported for them to find the role ARN and the file with the token.

This is usually handled in a seamless fashion when one uses the AWS SDK the expected way, but I see there's some custom configuration going on when setting up S3 (I assume to allow interoperability with services that have an S3-like API).

I added a check here for the environment variables EKS uses, and if they are found, setting up credentials using them. Since the order of credential providers matters, I also change the static provider setup to only prepend that provider to the chain when static credentials are configured.

### DIFF
--- a/pkg/new_storage/s3.go
+++ b/pkg/new_storage/s3.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"io"
 	"net/http"
+	"os"
 	"path"
 	"strings"
 	"time"
@@ -21,6 +22,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/new_storage/s3.go
```diff
@@ -59,15 +61,29 @@ func (s *S3) Connect() error {
...
+	}
+
+	awsRoleARN := os.Getenv("AWS_ROLE_ARN")
+	awsWebIdentityTokenFile := os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE")
+	if awsRoleARN != "" && awsWebIdentityTokenFile != "" {
+		cfg := &aws.Config{
+			Region: aws.String(s.Config.Region),
+		}
+		stsSvc := sts.New(session.Must(session.NewSession(cfg)))
+		stsProvider := stscreds.NewWebIdentityRoleProvider(stsSvc, awsRoleARN, "", awsWebIdentityTokenFile)
```
Slach, line 83: What is `""` magic constant here?

ldx: > What is `""` magic constant here?

Session name. When it's an empty string, the AWS SDK will use a timestamp based session name to generate a unique one.

pkg/new_storage/s3.go
```diff
@@ -59,15 +61,29 @@ func (s *S3) Connect() error {
...
+		// Define custom static cred provider
+		staticCreds := &credentials.StaticProvider{Value: credentials.Value{
+			AccessKeyID:     s.Config.AccessKey,
+			SecretAccessKey: s.Config.SecretKey,
+		}}
+		// Append static creds to the defaults
+		customCredProviders = append([]credentials.Provider{staticCreds}, customCredProviders...)
+	}
+
+	awsRoleARN := os.Getenv("AWS_ROLE_ARN")
```
Slach, line 76: Is these credentials stored only on enviroment variables or can be passed as config parameter?

ldx: > Is these credentials stored only on enviroment variables or can be passed as config parameter?

They are automatically added to the environment by Kubernetes.

Slach:
@ldx thanks a lot for contributing

PR #337, opened 2022-01-04
fix bug, recursuve=>recursive

wangzhen11aaa:
Words spell wrong.

### DIFF
--- a/pkg/new_storage/general.go
+++ b/pkg/new_storage/general.go
@@ -90,7 +90,7 @@ func isLegacyBackup(backupName string) (bool, string, string) {
 
 func (bd *BackupDestination) BackupList() ([]Backup, error) {
 	result := []Backup{}
-	err := bd.Walk("/", false, func(o RemoteFile) error {
+	err := bd.Walk("/", true, func(o RemoteFile) error {
 		// Legacy backup
 		if ok, backupName, fileExtension := isLegacyBackup(strings.TrimPrefix(o.Name(), "/")); ok {
 			result = append(result, Backup{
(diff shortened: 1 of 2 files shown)


wangzhen11aaa:
1. Recursive equals true is good for cos.
2.We can ignore some files with some other suffix.

Slach:
@wangzhen11aaa thanks a lot for contributing

pkg/new_storage/general.go
```diff
@@ -105,17 +105,14 @@ func (bd *BackupDestination) BackupList() ([]Backup, error) {
 			})
 			return nil
 		}
-		mf, err := bd.StatFile(path.Join(o.Name(), "metadata.json"))
+		metadataFilePath := ""
```
Slach, line 108: Did you run `./tests/integration/run.sh` to check your changes for other types of remote storage?

Slach: also, look like recursive walking will list lot of files which is not folder with backup name

Slach (changes requested):
look like your changes will increase resource usage for other types of remote storage

Slach:
@wangzhen11aaa integration tests failed
look details
https://github.com/AlexAkulov/clickhouse-backup/runs/4699579162

pkg/new_storage/general.go
wangzhen11aaa: > also, look like recursive walking will list lot of files which is not folder with backup name

I think that is ok. Without "recursive=true", we can not get some information about some backup.

wangzhen11aaa:
> @wangzhen11aaa integration tests failed look details https://github.com/AlexAkulov/clickhouse-backup/runs/4699579162

ok

PR #338, opened 2022-01-05
Support recursive

wangzhen11aaa:
fix bug: recursive should be true

### DIFF
--- a/pkg/new_storage/cos.go
+++ b/pkg/new_storage/cos.go
@@ -80,6 +80,8 @@ func (c *COS) DeleteFile(key string) error {
 func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) error) error {
 	prefix := path.Join(c.Config.Path, cosPath)
 	delimiter := ""
+	// We force recursive equals true.
+	recursive = true
 	if !recursive {
 		delimiter = "/"
 	}
--- a/pkg/new_storage/cos.go
+++ b/pkg/new_storage/cos.go
@@ -77,7 +77,7 @@ func (c *COS) DeleteFile(key string) error {
 	return err
 }
 
-func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) error) error {
+func (c *COS) doWalk(cosPath string, recursive bool, process func(RemoteFile) error) error {
 	prefix := path.Join(c.Config.Path, cosPath)
 	delimiter := ""
 	// We force recursive equals true.
@@ -110,6 +110,11 @@ func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) erro
 		}
 	}
 	return nil
+}
+
+func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) error) error {
+	// COS only supports recusive.
+	return c.doWalk(cosPath, true, process)
 }
 
 func (c *COS) GetFileReader(key string) (io.ReadCloser, error) {
--- a/pkg/new_storage/cos.go
+++ b/pkg/new_storage/cos.go
@@ -77,10 +77,10 @@ func (c *COS) DeleteFile(key string) error {
 	return err
 }
 
-func (c *COS) Walk(cosPath string, recursuve bool, process func(RemoteFile) error) error {
+func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) error) error {
 	prefix := path.Join(c.Config.Path, cosPath)
 	delimiter := ""
-	if !recursuve {
+	if !recursive {
 		delimiter = "/"
 	}
 	res, _, err := c.client.Bucket.Get(context.Background(), &cos.BucketGetOptions{
--- a/pkg/new_storage/s3.go
+++ b/pkg/new_storage/s3.go
@@ -203,6 +203,8 @@ func (s *S3) StatFile(key string) (RemoteFile, error) {
 }
 
 func (s *S3) Walk(s3Path string, recursive bool, process func(r RemoteFile) error) error {
+	// We force recursive equals true.
+	recursive = true
 	g, _ := errgroup.WithContext(context.Background())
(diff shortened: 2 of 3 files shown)


Slach:
@wangzhen11aaa failed again the same error
https://github.com/AlexAkulov/clickhouse-backup/runs/4711445590

wangzhen11aaa:
> @wangzhen11aaa failed again the same error https://github.com/AlexAkulov/clickhouse-backup/runs/4711445590

That's weird, I only changed the BackupList argument recursive from false to true, and this change should return more backup items which can contain the target backup. Right?

Slach:
> That's weird, I only changed the BackupList argument recursive from false to true, and this change should return more backup items which can contain the target backup. Right?

no, it just should walk via remote storage list with recursive order, and maybe it have side effects for other type of remote storage

you should try to figure out how to run e2e integration tests `./tests/integration/run.sh` (golang 1.17 + docker-compose + docker required for it)

wangzhen11aaa:
> 

Yes, I think this is a bug for other modules's design ,Which is a unexpected result.

pkg/new_storage/s3.go
```diff
@@ -203,6 +203,8 @@ func (s *S3) StatFile(key string) (RemoteFile, error) {
 }
 
 func (s *S3) Walk(s3Path string, recursive bool, process func(r RemoteFile) error) error {
+	// We force recursive equals true.
+	recursive = true
```
Slach: this change is suboptimal
try to figure out how exactly it works

Slach (changes requested):

Slach:
> Yes, I think this is a bug for other modules's design ,Which is a unexpected result.
No, it's not a bug for other modules, it works from beginning

You just don't figure out how `Walk` works

Unfortunately, I don't have ability now for COS testing in e2e ci/cd pipeline

wangzhen11aaa:
> 


> > Yes, I think this is a bug for other modules's design ,Which is a unexpected result.
> > No, it's not a bug for other modules, it works from beginning
> 
> You just don't figure out how `Walk` works
> 
> Unfortunately, I don't have ability now for COS testing in e2e ci/cd pipeline

I checked the ci test script, there is no COS test.

wangzhen11aaa:
> > Yes, I think this is a bug for other modules's design ,Which is a unexpected result.
> > No, it's not a bug for other modules, it works from beginning
> 
> You just don't figure out how `Walk` works
> 
> Unfortunately, I don't have ability now for COS testing in e2e ci/cd pipeline

I wonder how this COS added into project without CI test..

Slach:
> I wonder how this COS added into project without CI test..

It was added @AlexAkulov

pkg/new_storage/cos.go
```diff
@@ -77,10 +77,12 @@ func (c *COS) DeleteFile(key string) error {
 	return err
 }
 
-func (c *COS) Walk(cosPath string, recursuve bool, process func(RemoteFile) error) error {
+func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) error) error {
 	prefix := path.Join(c.Config.Path, cosPath)
 	delimiter := ""
-	if !recursuve {
+	// We force recursive equals true.
+	recursive = true
```
Slach, line 84: Maybe better will just always use `delimiter := "" `

And complete remove `recursive = true` and `if !recusive { delimiter = "/" }` ?

Slach (changes requested):

Slach:
> I checked the ci test script, there is no COS test.

yes, if you have valid COS account
look to GCP_TESTS in code
and try to add COS to `intergation_test.go`

pkg/new_storage/cos.go
wangzhen11aaa: > Maybe better will just always use `delimiter := "" `
> 
> And complete remove `recursive = true` and `if !recusive { delimiter = "/" }` ?

For the reason that Go does not support default value for parameters, and we can not remove this "recursive" because we don't want to break the interface.
Maybe, we can add another function that pass the true to next invocation. That's the rule of COS(Close for modification and Open for Extension) Rule. How about this?

wangzhen11aaa:
> intergation_test

Sorry, This account belongs to my company and has a high privilege. I don't have one.

wangzhen11aaa:
> 

I added an inner function (started with little word style).

Slach:
@wangzhen11aaa could you check latest docker version of with your `config.yml` and reports about all workflow (create + upload and download + restore) works for COS?
Something like that:

```
docker run -v /etc/clickhouse-backup/config.yml:/etc/clickhouse-backup/config.yml -v /var/lib/clickhouse:/var/lib/clickhouse --rm alexakulov/clickhouse-backup:master create test_backup

docker run -v /etc/clickhouse-backup/config.yml:/etc/clickhouse-backup/config.yml -v /var/lib/clickhouse:/var/lib/clickhouse --rm alexakulov/clickhouse-backup:master upload test_backup

docker run -v /etc/clickhouse-backup/config.yml:/etc/clickhouse-backup/config.yml -v /var/lib/clickhouse:/var/lib/clickhouse --rm alexakulov/clickhouse-backup:master delete local test_backup

docker run -v /etc/clickhouse-backup/config.yml:/etc/clickhouse-backup/config.yml -v /var/lib/clickhouse:/var/lib/clickhouse --rm alexakulov/clickhouse-backup:master download test_backup

docker run -v /etc/clickhouse-backup/config.yml:/etc/clickhouse-backup/config.yml -v /var/lib/clickhouse:/var/lib/clickhouse --rm alexakulov/clickhouse-backup:master restore --rm test_backup

docker run -v /etc/clickhouse-backup/config.yml:/etc/clickhouse-backup/config.yml -v /var/lib/clickhouse:/var/lib/clickhouse --rm alexakulov/clickhouse-backup:master delete remote test_backup

```

PR #341, opened 2022-01-06
Fix bug about metadata.json, we should check the file name first, ins…

wangzhen11aaa:
Hi, We should check whether does  "metadata.json" exist in file path first.

### DIFF
--- a/pkg/new_storage/general.go
+++ b/pkg/new_storage/general.go
@@ -88,6 +88,17 @@ func isLegacyBackup(backupName string) (bool, string, string) {
 	return false, backupName, ""
 }
 
+func getMetadataFilePath(file RemoteFile) string {
+	_metadataFilePath := ""
+
+	if strings.Contains(file.Name(), "metadata.json") {
+		_metadataFilePath = file.Name()
+	} else {
+		_metadataFilePath = path.Join(file.Name(), "metadata.json")
+	}
+	return _metadataFilePath
+}
+
 func (bd *BackupDestination) BackupList() ([]Backup, error) {
 	result := []Backup{}
 	err := bd.Walk("/", false, func(o RemoteFile) error {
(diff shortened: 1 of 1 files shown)


pkg/new_storage/general.go
```diff
@@ -88,6 +88,17 @@ func isLegacyBackup(backupName string) (bool, string, string) {
 	return false, backupName, ""
 }
 
+func getMetadataFilePath(file RemoteFile) string {
+	_metadataFilePath := ""
```
Slach, line 92: this is bad code style
please remove underscore

pkg/new_storage/general.go
```diff
@@ -135,7 +147,7 @@ func (bd *BackupDestination) BackupList() ([]Backup, error) {
 		if err != nil {
 			result = append(result, Backup{
 				metadata.BackupMetadata{
-					BackupName: strings.Trim(o.Name(), "/"),
+					BackupName: metadataFilePath,
```
Slach, line 150: looks like you broke 
`clickhouse-backup list remote`
for other storage types
you should set `backup_name` here
not `backup_name/metadata.json`

Looks like you previous fixes with delimiter and recursive was wrong
could you clarify why `Bucket.Get` from COS on `cos.go`  Walk with delimiter=/ for !recursive was not work properly?








functionality

pkg/new_storage/general.go
```diff
@@ -151,7 +163,7 @@ func (bd *BackupDestination) BackupList() ([]Backup, error) {
 		if err := json.Unmarshal(b, &m); err != nil {
 			result = append(result, Backup{
 				metadata.BackupMetadata{
-					BackupName: strings.Trim(o.Name(), "/"),
+					BackupName: metadataFilePath,
```
Slach, line 166: looks like you broke 
`clickhouse-backup list remote`
for other storage types
you should set `backup_name` here
not `backup_name/metadata.json`

Slach (changes requested):

pkg/new_storage/general.go
wangzhen11aaa: yes, I will check that. thanks.

pkg/new_storage/general.go
wangzhen11aaa: > looks like you broke `clickhouse-backup list remote` for other storage types you should set `backup_name` here not `backup_name/metadata.json`
> 
> Looks like you previous fixes with delimiter and recursive was wrong could you clarify why `Bucket.Get` from COS on `cos.go` Walk with delimiter=/ for !recursive was not work properly?
> 
> functionality

Yes

PR #342, opened 2022-01-06
Fix COS problem

wangzhen11aaa:
clickhouse-backup list remote

???        28/12/2021 05:47:22   remote      broken (can't stat metadata.json)
download_daily_mv_backup8  

->

download_daily_mv_backup-8  585B       06/01/2022 12:25:26   remote      tar

---------------------

### DIFF
--- a/pkg/new_storage/cos.go
+++ b/pkg/new_storage/cos.go
@@ -78,8 +78,23 @@ func (c *COS) DeleteFile(key string) error {
 }
 
 func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) error) error {
-	prefix := path.Join(c.Config.Path, cosPath)
+	// COS needs prefix ended with "/".
+	prefix := path.Join(c.Config.Path, cosPath) + "/"
+
 	delimiter := ""
+	if !recursive {
+		//
+		// When delimiter is "/", we only process all backups in the CommonPrefixes field of response.
+		// Then we get backupLists.
+		//
+		delimiter = "/"
+	} else {
+		//
+		// When delimiter is an empty string, we  process the items under specified path.
+		// Then we can Delete File object.
+		//
+		delimiter = ""
+	}
 	res, _, err := c.client.Bucket.Get(context.Background(), &cos.BucketGetOptions{
 		Delimiter: delimiter,
 		Prefix:    prefix,
@@ -87,6 +102,7 @@ func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) erro
 	if err != nil {
 		return err
 	}
+	// When recursive is false, only process all the backups in the CommonPrefixes part.
 	for _, dir := range res.CommonPrefixes {
 		if err := process(&cosFile{
 			name: strings.TrimPrefix(dir, prefix),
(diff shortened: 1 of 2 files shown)


pkg/new_storage/cos.go
```diff
@@ -78,30 +78,48 @@ func (c *COS) DeleteFile(key string) error {
 }
 
 func (c *COS) Walk(cosPath string, recursive bool, process func(RemoteFile) error) error {
-	prefix := path.Join(c.Config.Path, cosPath)
+	// COS needs prefix ended with "/".
+	prefix := path.Join(c.Config.Path, cosPath) + "/"
```
Slach, line 82: looks like you have wrong config
where `path` in `cos` section is empty
better will check this case or add config validation

pkg/new_storage/cos.go
```diff
@@ -78,30 +78,48 @@ func (c *COS) DeleteFile(key string) error {
...
 	}
-	for _, v := range res.Contents {
-		modifiedTime, _ := parseTime(v.LastModified)
-		if err := process(&cosFile{
-			name:         strings.TrimPrefix(v.Key, prefix),
-			lastModified: modifiedTime,
-			size:         int64(v.Size),
-		}); err != nil {
-			return err
+	if recursive {
```
Slach, line 113: looks much better

pkg/new_storage/general.go
```diff
@@ -135,7 +147,7 @@ func (bd *BackupDestination) BackupList() ([]Backup, error) {
 		if err != nil {
 			result = append(result, Backup{
 				metadata.BackupMetadata{
-					BackupName: strings.Trim(o.Name(), "/"),
+					BackupName: metadataFilePath,
```
Slach: still strange backup name, I propose return `strings.Trim(o.Name(), "/")` here

pkg/new_storage/general.go
```diff
@@ -151,7 +163,7 @@ func (bd *BackupDestination) BackupList() ([]Backup, error) {
 		if err := json.Unmarshal(b, &m); err != nil {
 			result = append(result, Backup{
 				metadata.BackupMetadata{
-					BackupName: strings.Trim(o.Name(), "/"),
+					BackupName: metadataFilePath,
```
Slach: still strange backup name, I propose return `strings.Trim(o.Name(), "/")` here

Slach (changes requested):

pkg/new_storage/cos.go
wangzhen11aaa: It is not ok. Passing bool value will turn function more complex. We can optimize this later. TODO, Split function with bool variable into two functions.

pkg/new_storage/general.go
wangzhen11aaa: The core logic to get the backup is through the metadata.json file, right? The name "metadataFilePath" variable here  makes this concept clear and conceals other messy result. If you insist, I can turn this code back, That's Ok

pkg/new_storage/cos.go
wangzhen11aaa: > looks like you have wrong config where `path` in `cos` section is empty better will check this case or add config validation

I use COS the first time, the path config is empty, It does not work. When check this API doc, u can find the "/" suffixed with every path in their query case. (%2F = "/", the html code).

wangzhen11aaa: > looks like you have wrong config where `path` in `cos` section is empty better will check this case or add config validation

I used COS the first time, the path config is exactly empty, It did not work either. When check this API doc, I can find the "/" suffixed with every path in their query case. (%2F = "/", the html code).

Slach (approved):

PR #345, opened 2022-01-07
Better Variable name for its function

wangzhen11aaa:
Does new name "DuplicatedBackup" reflect the meaning of the function than "Required"?

### DIFF
--- a/pkg/metadata/metadata.go
+++ b/pkg/metadata/metadata.go
@@ -51,9 +51,9 @@ type TableMetadata struct {
 }
 
 type Part struct {
-	Partition string `json:"partition,omitempty"`
-	Name      string `json:"name"`
-	Required  bool   `json:"required,omitempty"`
+	Partition      string `json:"partition,omitempty"`
+	Name           string `json:"name"`
+	DuplicatedPart bool   `json:"required,omitempty"`
 	// Path                              string    `json:"path"`              // TODO: должен быть относительный путь вообще непонятно зачем он, его можно из name получить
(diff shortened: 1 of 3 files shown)


pkg/metadata/metadata.go
```diff
@@ -51,9 +51,9 @@ type TableMetadata struct {
 }
 
 type Part struct {
-	Partition string `json:"partition,omitempty"`
-	Name      string `json:"name"`
-	Required  bool   `json:"required,omitempty"`
+	Partition      string `json:"partition,omitempty"`
+	Name           string `json:"name"`
+	DuplicatedPart bool   `json:"required,omitempty"`
```
Slach, line 56: Sorry, I can't approve this PR
I undestand maybe  "requiered" is not so mental name as you expected in the code, but differences between `json` and internal variable name is worst variant.

Slach (commented):
Sorry, I can't approve this PR
I undestand maybe "requiered" is not so mental name as you expected in the code, but differences between json and internal variable name is worst variant.

pkg/metadata/metadata.go
wangzhen11aaa: > Sorry, I can't approve this PR I undestand maybe "requiered" is not so mental name as you expected in the code, but differences between `json` and internal variable name is worst variant.

What does this "required" mean?

Slach: For incremental backup, we have the parent backup which marked as `required` in backup_name/metadata.json
and some parts are not present in incremental backup and present only in `required` backup, so it also marked as `required`

During download incremental backup, `required` backup downloads first and during unpack `required` parts create as hard links.

PR #355, opened 2022-01-15
Code should be placed in proper position

wangzhen11aaa:


pkg/clickhouse/clickhouse.go
```diff
@@ -705,9 +571,76 @@ func IsClickhouseShadow(path string) bool {
...
+		}
+
+		if isPtr {
+			direct.Set(reflect.Append(direct, vp))
+		} else {
+			direct.Set(reflect.Append(direct, v))
+		}
+
+		// rt := reflect.TypeOf(s)
+		// if rt.Kind() != reflect.Struct {
```
Slach: let's remove this part of unused commented obslete code

pkg/clickhouse/flashback.go
```diff
@@ -68,7 +68,7 @@ package clickhouse
 // 	if err != nil {
 // 		return err
 // 	}
-// 	detachedParentDir := filepath.Join(dataPath, "data", TablePathEncode(diff.BTable.Database), TablePathEncode(diff.BTable.Table), "detached")
+// 	detachedParentDir := filepath.Join(dataPath, "data", common.TablePathEncode(diff.BTable.Database), common.TablePathEncode(diff.BTable.Table), "detached")
```
Slach, line 71: If you will figure out how flashback command works
and add test for flashback command to `./test/integration/integration_test.go`
it will great

in feature PR, if you found unused commented code part, feel free to remove it

pkg/common/utils.go
```diff
@@ -0,0 +1,11 @@
+package common
+
+import (
+	"net/url"
+	"strings"
+)
+
+func TablePathEncode(str string) string {
+	return strings.ReplaceAll(
+		strings.ReplaceAll(url.PathEscape(str), ".", "%2E"), "-", "%2D")
```
Slach: you can use   here
```golang
return strings.NewReplacer(".", "%2E", "-", "%2D").Replace(url.PathEscape(str))
```

Slach (changes requested):

pkg/common/utils.go
wangzhen11aaa: > you can use here
> 
> ```go
> return strings.NewReplacer(".", "%2E", "-", "%2D").Replace(url.PathEscape(str))
> ```

We can extract this as function and place it in the common package.

pkg/clickhouse/flashback.go
wangzhen11aaa: > If you will figure out how flashback command works and add test for flashback command to `./test/integration/integration_test.go` it will great
> 
> in feature PR, if you found unused commented code part, feel free to remove it

For the merging reason, I don't think flashback works for mergetree. Hashes can not work out, what's more I do not know much about this function.

pkg/common/utils.go
Slach: resolve this comment

Slach (approved):

Slach:
Thanks a lot for contribution, I hope future PRs will improve functionality instead of just refactoring "as you like"

PR #407, opened 2022-03-01
Fix trimming of S3_PATH prefix when walking S3 bucket content

### LINKED ISSUE #406: S3_PATH is not cleaned correctly when listing remote backups
Due to how `clickhouse-backup` interacts with the S3 API when listing existing backups, the `S3_PATH` is joined twice to the backup name before fetching the `metadata.json` file resulting in something like:

```HTTP
HEAD /<bucket>/<prefix>/<prefix>/<backup-name>/metadata.json
```

This isn't an issue as long as the prefix is empty but if you try to create backups in sub-directories it's not possible to list/restore them later on.

I'd be glad to provide a fix if desired :blush:

prskr:
Closes #406

### DIFF
--- a/pkg/new_storage/s3.go
+++ b/pkg/new_storage/s3.go
@@ -3,13 +3,14 @@ package new_storage
 import (
 	"context"
 	"crypto/tls"
-	"github.com/AlexAkulov/clickhouse-backup/pkg/config"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


Slach:
list remote doesn't work after your changes
could you share your `clickhouse-backup print-config`?

PR #408, opened 2022-03-03
fix api list table

benbiti:
showCounts not changed, make only index 0 updated, other items is empty.


Slach:
good catch, thanks a lot for your contribution

PR #414, opened 2022-03-16
Fix incorrect in progress check on the example of Kubernetes CronJob

hgiasac:
The status should be `in progress`, not `InProgress`


Slach (approved):

Slach:
good catch, thanks!

PR #459, opened 2022-06-22
Add FreezeByPartWhere config to filter partition

vahid-sohrabloo:
Sometimes we need to filter partitions for backup. like we don't want to backup parts if they are on s3.


Slach:
@vahid-sohrabloo hello thanks for contributing, could you add your config option description in `ReadMe.md`?
Moreover, adding separate test case `tests/integration/integration_test.go` will useful

Issue #495, opened 2022-08-15
API Server add an http api for updating configuration files?


Fkylwj:
I need to use clickhouse-backup as an http api server to back up data on the ftp server. 
Since the ftp-related configuration may need to be changed, can I submit a PR to add an http interface to update the configuration file?

Slach:
`clickhouse-backup server` reload config every time when you try to execute command
so just change your config on the disk and it will apply automatically when next command will executed via API

PR #500, opened 2022-08-18
feat: api server add an http interface to support update config

Fkylwj:
issue #495

Sometimes need to change config file via http api when using clickhouse-backup as api server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,24 @@
-FROM clickhouse/clickhouse-server:latest
+FROM hub.deepin.com/library/golang:bullseye AS builder
+WORKDIR /src
+ADD . .
+RUN go env -w GOPROXY="https://goproxy.cn,direct"
+RUN make config
+
+
+FROM hub.deepin.com/public/alpine:3.16.0-udcp AS runner
 ARG TARGETPLATFORM
 
-RUN wget -qO- https://kopia.io/signing-key | gpg --dearmor -o /usr/share/keyrings/kopia-keyring.gpg && \
-    echo "deb [signed-by=/usr/share/keyrings/kopia-keyring.gpg] http://packages.kopia.io/apt/ stable main" > /etc/apt/sources.list.d/kopia.list && \
-    apt-get update -y && \
-    apt-get install -y ca-certificates tzdata bash curl restic rsync rclone jq gpg && \
-    update-ca-certificates && \
-    rm -rf /var/lib/apt/lists/* && rm -rf /var/cache/apt/*
+RUN addgroup -g 101 -S clickhouse && adduser -S -h /var/lib/clickhouse -s /bin/bash -G clickhouse -g "ClickHouse server" -u 101 clickhouse # buildkit
+RUN apk update && apk add --no-cache ca-certificates tzdata bash curl && update-ca-certificates # buildkit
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,19 +1,24 @@
-FROM clickhouse/clickhouse-server:latest
+FROM hub.deepin.com/library/golang:bullseye AS builder
```
Slach, line 1: sorry, we can't approve these changes
could you use `FROM golang:1.19 AS builder`  here?

Dockerfile
```diff
@@ -1,19 +1,24 @@
...
+WORKDIR /src
+ADD . .
+RUN go env -w GOPROXY="https://goproxy.cn,direct"
```
Slach, line 4: sorry, it's a China specific changes lot of users can't use goproxy.cn cause it have high latency

Dockerfile
```diff
@@ -1,19 +1,24 @@
...
+RUN make config
+
+
+FROM hub.deepin.com/public/alpine:3.16.0-udcp AS runner
```
Slach, line 8: we decided to switch to ubuntu based platform to allow install some thrird paty extenal tools which could integrate for `remote_storage: custom`

Dockerfile
```diff
@@ -1,19 +1,24 @@
...
 
 COPY entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
-COPY build/${TARGETPLATFORM}/clickhouse-backup /bin/clickhouse-backup
+COPY --from=builder /src/build/${TARGETPLATFORM}/clickhouse-backup /bin/clickhouse-backup
+COPY --from=builder /src/clickhouse-backup/config.yml /etc/clickhouse-backup/config.yml
 RUN chmod +x /bin/clickhouse-backup
 
-# USER clickhouse
+#docker 容器运行，clickhouse数据存储路径利用数据卷挂载的方式，不能单独指定clickhouse用户
```
Slach, line 20: Please, use only English language

Makefile
```diff
@@ -101,3 +101,18 @@ build-race: $(NAME)/$(NAME)-race
 
 $(NAME)/$(NAME)-race: $(GO_FILES)
 	CGO_ENABLED=1 $(GO_BUILD) -gcflags "all=-N -l" -race -o $@ ./cmd/$(NAME)
+
+
+# 仓库地址基础路径  默认hub.deepin.com/wuhan_udcp
```
Slach, line 106: please use English

Makefile
```diff
@@ -101,3 +101,18 @@ build-race: $(NAME)/$(NAME)-race
...
+DOCKER_BASE := $(if $(DOCKER_BASE),$(DOCKER_BASE:%/=%),hub.deepin.com)/wuhan_udcp
```
Slach, line 107: we can't use extenral image hare

Slach (changes requested):
I don't like idea update config during
it looks like security hole
by default anybody can update backup config and re-define backup destination
for dump whole database

Slach:
unfortunately we can't approve these functionality by security reason

i think more properly way change config.yml via ansible / puppet / kubernetes configmap
and changed config will load and apply to next API command execution

PR #556, opened 2022-11-04
fix deleting backups for S3 with versioning

### LINKED ISSUE #555: S3 backup directories are not deleted if versioning is enabled
Hi
We faced the problem with remaining empty directories for deleted backup from S3. We use Minio with enabled replication. Versioning is also enabled for buckets. We assume that backup folder cannot be deleted properly because version Id must be provided [here](https://github.com/AlexAkulov/clickhouse-backup/blob/master/pkg/storage/s3.go#L191) in case of versioning enabled.

Tvion:
fix https://github.com/AlexAkulov/clickhouse-backup/issues/555


Slach:
thank you so much for your contribution

Issue #568, opened 2022-11-24
watch cmd fail on time - error sql: database is closed


nestyurkin:
cmd to run
`clickhouse-backup -c /etc/clickhouse-backup/config-test.yml watch --tables=dwh_meta.*`
VERSION:
   2.1.2
last log before quit
```
2022/11/24 16:53:26.324808  info clickhouse connection open: tcp://localhost:9005 logger=clickhouse
2022/11/24 16:53:26.324832  info SELECT * FROM system.disks; logger=clickhouse
[clickhouse]host(s)=localhost:9005, database=system, username=default
[clickhouse][dial] secure=false, skip_verify=false, strategy=random, ident=45, server=0 -> 127.0.0.1:9005
[clickhouse][connect=45][hello] -> Golang SQLDriver 1.1.54213
[clickhouse][connect=45][hello] <- ClickHouse 22.10.54460 (Europe/Moscow)
[clickhouse][connect=45][prepare] SELECT * FROM system.disks;
[clickhouse][connect=45][send query] SELECT * FROM system.disks;
[clickhouse][connect=45][query settings] connect_timeout=900&receive_timeout=900&send_timeout=900
[clickhouse][connect=45][send external tables] count 0
[clickhouse][connect=45][read meta] <- data: packet=1, columns=9, rows=0
[clickhouse][connect=45][rows] <- data: packet=1, columns=9, rows=1, elapsed=18.873µs
[clickhouse][connect=45][rows] <- profiling: rows=1, bytes=37760, blocks=1
[clickhouse][connect=45][rows] <- progress: rows=1, bytes=93, total rows=0
[clickhouse][connect=45][rows] <- data: packet=1, columns=0, rows=0, elapsed=750ns
[clickhouse][connect=45][rows] <- progress: rows=0, bytes=0, total rows=0
[clickhouse][connect=45][rows] <- end of stream
[clickhouse][connect=45][rows] close
[clickhouse][connect=45][stmt] close
2022/11/24 16:53:26.327705 debug remove '/clickhouse/backup/test-full-20221124135314' logger=RemoveBackupLocal
2022/11/24 16:53:26.329460  info done                      backup=test-full-20221124135314 duration=7ms location=local logger=RemoveBackupLocal operation=delete
2022/11/24 16:53:26.329487  info clickhouse connection closed logger=clickhouse
[...]

PR #569, opened 2022-11-24
Watch interval validation fix

clamoris:
Current validation prevents from properly using watch functionality as it requires a wrong number of stored backups to work.

Example:
1h watch interval, 5h full interval needs at least 5 stored remote backups to work,  but validation only accepts 4 or less.

### DIFF
--- a/pkg/backup/watch.go
+++ b/pkg/backup/watch.go
@@ -53,7 +53,7 @@ func (b *Backuper) ValidateWatchParams(watchInterval, fullInterval, watchBackupN
 	if watchBackupNameTemplate != "" {
 		b.cfg.General.WatchBackupNameTemplate = watchBackupNameTemplate
 	}
-	if b.cfg.General.FullDuration.Seconds() < b.cfg.General.WatchDuration.Seconds()*float64(b.cfg.General.BackupsToKeepRemote) {
+	if b.cfg.General.FullDuration.Seconds() > b.cfg.General.WatchDuration.Seconds()*float64(b.cfg.General.BackupsToKeepRemote) {
 		return fmt.Errorf("fullInterval `%s` is not enought to keep %d remote backups with watchInterval `%s`", b.cfg.General.FullInterval, b.cfg.General.BackupsToKeepRemote, b.cfg.General.WatchInterval)
 	}
 	return nil
--- a/Examples.md
+++ b/Examples.md
@@ -378,13 +378,13 @@ spec:
 
 ## How do incremental backups work to remote storage
 - Incremental backup calculate increment only during execute `upload` or `create_remote` command or similar REST API request.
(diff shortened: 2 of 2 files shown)


clamoris:
On the side note --watch-backup-name-template seems to have some problems too.

Pattern "ch_{replica}-{type}-{time:20060102150405}" results in default `shard01-full-20221124155826` name. `replica` is a macros.

pkg/backup/watch.go
```diff
@@ -53,7 +53,7 @@ func (b *Backuper) ValidateWatchParams(watchInterval, fullInterval, watchBackupN
 	if watchBackupNameTemplate != "" {
 		b.cfg.General.WatchBackupNameTemplate = watchBackupNameTemplate
 	}
-	if b.cfg.General.FullDuration.Seconds() < b.cfg.General.WatchDuration.Seconds()*float64(b.cfg.General.BackupsToKeepRemote) {
+	if b.cfg.General.FullDuration.Seconds() > b.cfg.General.WatchDuration.Seconds()*float64(b.cfg.General.BackupsToKeepRemote) {
```
Slach, line 56: are you sure this will help resolve your "connection close" issue?

Slach:
oops sorry, miss you with author of https://github.com/AlexAkulov/clickhouse-backup/issues/568

pkg/backup/watch.go
Slach: Sorry, I don't understand what exactly you try to fix
this part of code is not related to `current validation allow backups_to_keep_remote less than 4`

could you provide more examples in comments?

Slach (changes requested):

pkg/backup/watch.go
clamoris: Sure.

Current (wrong) version in pseudo-code:
```
If FullDurationSeconds < (BackupsToKeep * WatchInterval) {
    raise validation error
}
```

It prevents user from increasing amount of backups - if we set BackupsToKeep to infinity validation starts failing. 

Example:
Let say we want to make incremental daily backups and have full backup on Sunday and to store two weeks of backups on remote.

```
FullDurationSeconds = 1 week (604800 seconds)
BackupsToKeep = 14
WatchInterval = 1 day (86400 seconds)
```

Validation will fail, since `604800 < (14 * 86400)`, it would only work if `BackupsToKeep` is less than 7, which is not enough to store a full week. With the changes it will accept any value of `BackupsToKeep` larger than 7.

Slach: ok. got it
could you fix intergration_test.go for https://github.com/AlexAkulov/clickhouse-backup/actions/runs/3541456947/jobs/5949819982#step:7:56324

look like your changes broke some backup API scenarios

Slach: @clamoris any progress?

clamoris: Tests on my machine are ridiculously slow, something is not right
Screenshot 2022-12-02 at 21 52 45

PR #626, opened 2023-03-28
Readme and Manual file improvements

o-mdr:
Updated main.go with minor typo fixes. Regenerated manual markdown file. Removed contents of the manual from Readme file (because it is exact duplication, readme file is long enough, and it's easier to maintain just one source of the manual).

The following has changed:
- Minor English typo corrections
- Improvements to commands documentation


ReadMe.md
```diff
@@ -65,44 +65,47 @@ DESCRIPTION:
    Run as 'root' or 'clickhouse' user
 
 COMMANDS:
-   tables               List list of tables, exclude skip_tables
+   tables               List of tables, exclude skip_tables
```
Slach: This part of README is generated
from `cmd/clickhouse-backup/main.go`
please change it instead

Slach (changes requested):

ReadMe.md
Slach: also look to https://github.com/Altinity/clickhouse-backup/blob/master/generate_manual.sh
which Generate Manual.md
which copy and pasted to Readme.md

o-mdr: Oh I see how this works. I've regenerated file, moved contents to just Manual.md (rather than keep it in Readme and Manual). I've left a link in the readme file to point to the manual.

Slach (approved):

PR #636, opened 2023-04-18
server: Add callbacks query param

### LINKED ISSUE #635: Async API Callback
Would it be OK to make it so that you can pass an async API callback to the clickhouse-backup server, so that the caller can be notified when some operation is complete?

mskwon:
Add callbacks query param, which can be used by clients to the server for notifications on completion of an asynchronous operation.

Callback support is added to the following endpoints:

    /backup/create
    /backup/upload/{name}
    /backup/download/{name}
    /backup/restore/{name}

The callback parameter can be passed to the target endpoint with the addition of

?callback={URI Encoded Callback URL}

Callback payloads are POSTed with the JSON serialized struct CallbackResponse, which is exported for use by clients. Multiple callback URLs can be specified; all URLs will be hit regardless of failures in any single endpoint.

Resolves AlexAkulov/clickhouse-backup#635

### DIFF
--- a/pkg/server/callback.go
+++ b/pkg/server/callback.go
@@ -1,0 +1,64 @@
+package server
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+// callbackFn is a function which will post a callback when invoked
+type callbackFn func(ctx context.Context, v interface{}) []error
+
+// parseCallback parses a callback URL from URL query values and returns a closure which can send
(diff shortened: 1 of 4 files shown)


Slach:
looks amazing, thanks for contribution, need more time to review

Slach (changes requested):
code looks clear
could you add callback parameter description to ReadMe.md API sections?

Slach (approved):

Slach:
i will add it later

mskwon:
Thank you!

PR #655, opened 2023-05-17
proposal: ensure only one full backup will be created

farbodsalimi:
Currently, the cron job example creates several full backups on Mondays (FULL_BACKUP_WEEKDAY=1) if you schedule it to run more than one time a day. I propose ensuring only one full backup to be made on FULL_BACKUP_WEEKDAY.

### DIFF
--- a/Examples.md
+++ b/Examples.md
@@ -333,7 +333,7 @@ spec:
                   for SERVER in $CLICKHOUSE_SERVICES; do
                     if [[ "1" == "$MAKE_INCREMENT_BACKUP" ]]; then
                       LAST_FULL_BACKUP=$(clickhouse-client -q "SELECT name FROM system.backup_list WHERE location='remote' AND name LIKE '%full%' AND desc NOT LIKE 'broken%' ORDER BY created DESC LIMIT 1 FORMAT TabSeparatedRaw" --host="$SERVER" --port="$CLICKHOUSE_PORT" --user="$BACKUP_USER" $BACKUP_PASSWORD);
-                      TODAY_FULL_BACKUP=$(clickhouse-client -q "SELECT name FROM system.backup_list WHERE location='remote' AND name LIKE '%full%' AND desc NOT LIKE 'broken%' AND DATE_FORMAT(created, '%Y-%m-%d')=DATE_FORMAT(now(), '%Y-%m-%d') ORDER BY created DESC LIMIT 1 FORMAT TabSeparatedRaw" --host="$SERVER" --port="$CLICKHOUSE_PORT" --user="$BACKUP_USER" $BACKUP_PASSWORD)
+                      TODAY_FULL_BACKUP=$(clickhouse-client -q "SELECT name FROM system.backup_list WHERE location='remote' AND name LIKE '%full%' AND desc NOT LIKE 'broken%' AND toDate(created) = today() ORDER BY created DESC LIMIT 1 FORMAT TabSeparatedRaw" --host="$SERVER" --port="$CLICKHOUSE_PORT" --user="$BACKUP_USER" $BACKUP_PASSWORD)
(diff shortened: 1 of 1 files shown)


Examples.md
```diff
@@ -333,16 +333,17 @@ spec:
                   for SERVER in $CLICKHOUSE_SERVICES; do
                     if [[ "1" == "$MAKE_INCREMENT_BACKUP" ]]; then
                       LAST_FULL_BACKUP=$(clickhouse-client -q "SELECT name FROM system.backup_list WHERE location='remote' AND name LIKE '%full%' AND desc NOT LIKE 'broken%' ORDER BY created DESC LIMIT 1 FORMAT TabSeparatedRaw" --host="$SERVER" --port="$CLICKHOUSE_PORT" --user="$BACKUP_USER" $BACKUP_PASSWORD);
+                      TODAY_FULL_BACKUP=$(clickhouse-client -q "SELECT name FROM system.backup_list WHERE location='remote' AND name LIKE '%full%' AND desc NOT LIKE 'broken%' AND DATE_FORMAT(created, '%Y-%m-%d')=DATE_FORMAT(now(), '%Y-%m-%d') ORDER BY created DESC LIMIT 1 FORMAT TabSeparatedRaw" --host="$SERVER" --port="$CLICKHOUSE_PORT" --user="$BACKUP_USER" $BACKUP_PASSWORD)
```
Slach: Using `toDate(created) = today()` instead of `DATE_FORMAT` will shorter

Examples.md
```diff
@@ -333,16 +333,17 @@ spec:
...
                       PREV_BACKUP_NAME=$(clickhouse-client -q "SELECT name FROM system.backup_list WHERE location='remote' AND desc NOT LIKE 'broken%' ORDER BY created DESC LIMIT 1 FORMAT TabSeparatedRaw" --host="$SERVER" --port="$CLICKHOUSE_PORT" --user="$BACKUP_USER" $BACKUP_PASSWORD);
                       DIFF_FROM[$SERVER]="";
-                      if [[ "$FULL_BACKUP_WEEKDAY" == "$(date +%u)" || "" == "$PREV_BACKUP_NAME" || "" == "$LAST_FULL_BACKUP" ]]; then
+                      if [[ ("$FULL_BACKUP_WEEKDAY" == "$(date +%u)" && "" == "$TODAY_FULL_BACKUP") || "" == "$PREV_BACKUP_NAME" || "" == "$LAST_FULL_BACKUP" ]]; then
```
Slach, line 339: Could you please clarify, which corner case you are trying to avoid here?

If I understood correctly, you tried to change `CronJob` schedule to run more than one time per day? And FULL backup created multiples time on Monday or other case?

Slach (changes requested):

Slach:
thank you for contribution!

Examples.md
farbodsalimi: Yes exactly. I'm not sure how many people need to run their backups more than once a day but for example for some of our clusters, we need to run the backup corn job every hour which means with the current example on Mondays the jobs create 24 full backups.

Slach: This requires to change approach
instead of `"$FULL_BACKUP_WEEKDAY" == "$(date +%u)" `
something like  ` "$FULL_BACKUP_HOUR" == "$(date +%H)" `

Slach (approved):

Slach (approved):

Examples.md
farbodsalimi: I see yes there might be a better way to do that via the `date` command. IRL we want to be able to replicate what we had with the `watch` cmd:

```bash
clickhouse-backup watch --watch-interval=1h --full-interval=168h
```

1 incremental per hour
1 full per week

Slach: You can't run `clickhouse-backup watch` outside pod with clickhouse data, in this case you will back up only schema because `clickhouse-backup` requires access to the same data files as `clickhouse-server`

PR #670, opened 2023-06-14
[WIP] try switch to Zerolog

### LINKED ISSUE #624: Migrate from apexlog to zerolog, to avoid race-conditions

Slach:
I don't understand how use `zerolog` for multiple go-routines ;(

look details https://github.com/rs/zerolog/issues/555

fix https://github.com/Altinity/clickhouse-backup/issues/624


pkg/clickhouse/clickhouse.go
```diff
@@ -105,23 +105,23 @@ func (ch *ClickHouse) Connect() error {
 	}
 
 	if ch.conn, err = clickhouse.Open(opt); err != nil {
-		ch.Log.Errorf("clickhouse connection: %s, sql.Open return error: %v", fmt.Sprintf("tcp://%v:%v", ch.Config.Host, ch.Config.Port), err)
+		log.Error().Msgf("clickhouse connection: %s, clickhouse.Open return error: %v", fmt.Sprintf("tcp://%v:%v", ch.Config.Host, ch.Config.Port), err)
 		return err
 	}
 
-	logFunc := ch.Log.Infof
+	logFunc := log.Info()
```
rdmrcv: If a race condition still an issue, this line looks like a root of all problems.

When you call the `log.Info` function, you create a `zerolog.Event` struct. Not a function.

But the struct is pooled, and will be re-used later. BUT. Here you cache that struct. 

And then you call a `Msgf` method which will release a pointer to the struct to the pool. Each call of `Msgf` will put the pointer to the pool. (to be clear docs say that after first use of `Msg*` method the Event should be abandoned).

And you end up with 2 pointers to the same struct in the pool.

Then when you'll use that poisoned pool it had a chance to return same pointer twice and introduce a data-race.

Slach: oh, yep, issue still present, thanks a lot for review

what is workaround?

`logFunc := log.Info`
instead of `log.Info()`?

rdmrcv: Nope. You'll still point to same object twice.

Looks like correct workaround here should be something like that:

```go
logLevel := zerolog.InfoLevel

...
if (...) {
    logLevel = zerolog.ErrorLevel
}
...

log.WithLevel(logLevel).Msgf(...)
```

Then you'll receive new `Event` each time you call `WithLevel` but maintain loglevel separately as you want it to do.

Slach: @rdmrcv really thanks for help. I will try

rdmrcv: On the other hand, your solution should work too, I'm little bit overthinking here.

Each time you call `log.Info` you also receive new Event, not cache the same. And when you save the `log.Info` (not `log.Info()`) into a variable you actually cache pointer to `log.Info` and not result of its work. So it should work too.

rdmrcv: @Slach don't you have a chance to try this patch? Is it fix anything or behavior stays same? :)

Slach: @rdmrcv thanks you so much, for your suggestion, now all works, and test passed

PR #740, opened 2023-09-05
respect max_parts_count

### LINKED ISSUE #739: S3 max_parts_count not respected may result in TotalPartsExceeded
Because of the way the S3 part size is computed for a multipart upload, the resulting number of parts can exceed the value specified in max_parts_count.

In our case, max_parts_count is set to 10,000 (the maximum S3 value) and max_file_size is set to 1 TiB. When the backup size of a table reaches the max_file_size for uploading, the computed number of parts is more than 10,000 (I suspect it is 10,001) and the upload of the backup fails:
```
2023/09/01 06:52:24.444085 error UploadCompressedStream return error: MultipartUpload: upload multipart failed
upload id: LdghgFb82TZzmCc...
caused by: TotalPartsExceeded: exceeded total allowed configured MaxUploadParts (10000). Adjust PartSize to fit in this limit logger=uploadTableData`
```

config.yml
```
general:
  max_file_size: 1099511627776
  remote_storage: s3
  disable_progress_bar: True
  backups_to_keep_local: 2
  backups_to_keep_remote: 0
  log_level: error
  allow_empty_backups: False
  upload_by_part: False
  download_by_part: False
  download_concurrency: 32
  upload_concurrency: 16
clickhouse:
  username: default
  password:
  host: localhost
  port: 9000
  data_path: /var/lib/clickhouse
  skip_tables: ['system.*', 'information_schema.*', 'INFORMATION_SCHEMA.*']
  timeout: 5m
  freeze_by_part: False
  sync_replicated_tables: False
  log_sql_queries: False
  config_dir: /etc/clickhouse-server
  restart_command: systemctl restart clickhouse-server
s3:
  access_key:
  secret_key:
  bucket: backups-...-us-east-1-main
  endpoint:
  region: us-east-1
  acl: private
  force_path_style: False
  path: clickhouse
  disable_ssl: False
  compression_level: 1
  compression_format: tar
  sse: AES256
  disable_cert_verification: False
  storage_class: STANDARD
  part_size: 0
  max_parts_count: 10000
  debug: False
  concurrency: 16
  allow_multipart_download: True
``` 

As a workaround, we have decreased the max_parts_count to 5,000 to avoid exceeding the S3 maximum.

rodrigargar:
Fix https://github.com/Altinity/clickhouse-backup/issues/739

### DIFF
--- a/pkg/storage/general.go
+++ b/pkg/storage/general.go
@@ -633,6 +633,9 @@ func NewBackupDestination(ctx context.Context, cfg *config.Config, ch *clickhous
 		// https://github.com/Altinity/clickhouse-backup/issues/317
 		if bufferSize <= 0 {
 			bufferSize = int(cfg.General.MaxFileSize) / cfg.AzureBlob.MaxPartsCount
+			if int(cfg.General.MaxFileSize) % cfg.AzureBlob.MaxPartsCount > 0 {
+				bufferSize++
+			}
 			if bufferSize < 2*1024*1024 {
 				bufferSize = 2 * 1024 * 1024
 			}
@@ -652,6 +655,9 @@ func NewBackupDestination(ctx context.Context, cfg *config.Config, ch *clickhous
 		partSize := cfg.S3.PartSize
 		if cfg.S3.PartSize <= 0 {
 			partSize = cfg.General.MaxFileSize / cfg.S3.MaxPartsCount
+			if cfg.General.MaxFileSize % cfg.S3.MaxPartsCount > 0 {
+				partSize++
+			}
 			if partSize < 5*1024*1024 {
 				partSize = 5 * 1024 * 1024
 			}
--- a/pkg/storage/s3.go
+++ b/pkg/storage/s3.go
@@ -531,8 +531,11 @@ func (s *S3) CopyObject(ctx context.Context, srcBucket, srcKey, dstKey string) (
 
 	// Set the part size (e.g., 5 MB)
 	partSize := srcSize / s.Config.MaxPartsCount
+	if srcSize % s.Config.MaxPartsCount > 0 {
+		partSize++
+	}
 	if partSize < 5*1024*1024 {
-		partSize = 5 * 1014 * 1024
+		partSize = 5 * 1024 * 1024
 	}
 
 	// Calculate the number of parts


pkg/storage/general.go
```diff
@@ -633,6 +633,9 @@ func NewBackupDestination(ctx context.Context, cfg *config.Config, ch *clickhous
 		// https://github.com/Altinity/clickhouse-backup/issues/317
 		if bufferSize <= 0 {
 			bufferSize = int(cfg.General.MaxFileSize) / cfg.AzureBlob.MaxPartsCount
+			if int(cfg.General.MaxFileSize) % cfg.AzureBlob.MaxPartsCount > 0 {
+				bufferSize++
```
Slach, line 637: this changes do nothing

pkg/storage/s3.go
```diff
@@ -531,8 +531,11 @@ func (s *S3) CopyObject(ctx context.Context, srcBucket, srcKey, dstKey string) (
 
 	// Set the part size (e.g., 5 MB)
 	partSize := srcSize / s.Config.MaxPartsCount
+	if srcSize % s.Config.MaxPartsCount > 0 {
+		partSize++
+	}
 	if partSize < 5*1024*1024 {
-		partSize = 5 * 1014 * 1024
+		partSize = 5 * 1024 * 1024
```
Slach, line 538: good catch

pkg/storage/s3.go
```diff
@@ -531,8 +531,11 @@ func (s *S3) CopyObject(ctx context.Context, srcBucket, srcKey, dstKey string) (
 
 	// Set the part size (e.g., 5 MB)
 	partSize := srcSize / s.Config.MaxPartsCount
+	if srcSize % s.Config.MaxPartsCount > 0 {
+		partSize++
```
Slach, line 535: try to explain what exactly this change should fix?

Slach (changes requested):

pkg/storage/s3.go
rodrigargar: If the size of the upload is not divisible by MaxPartsCount, you will need MaxPartsCount + 1 parts to complete the upload with the calculated part size because the integer division truncates the result.

A couple of lines later:
```
// Calculate the number of parts (equals to MaxPartsCount + 1 if srcSize not divisible by MaxPartsCount)
numParts := (srcSize + partSize - 1) / partSize
```

 Adding 1 to the partSize ensures that the MaxPartsCount will not be exceeded.

Slach (approved):

Slach:
could you merge latest master to your branch?

PR #776, opened 2023-11-02
mention s3:GetBucketVersioning permission in README

vitaliyf:
https://github.com/Altinity/clickhouse-backup/pull/556 added a check if S3 bucket versioning is enabled, but this requires its own IAM permission.

Without this permissions there are periodic `AccessDenied` errors in logs while making `GET /?versioning= HTTP/1.1` from isVersioningEnabled() though they appear to be harmless if versioning isn't being used.


Slach:
nice catch, thanks for contributing

PR #784, opened 2023-11-14
Improve GCP uploading

minguyen9988:
thanks Andrew W for the contribution.

The PR include 2 fix:
- GCP upload need to return the error, currently it cannot catch upload error and retry, because the error is handle in defer and never return to caller, so it will fail at the first failure.
- Add support for HTTP reverse proxy handling, if the flag force_http is set to true it will use http reverse proxy.

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -509,6 +509,7 @@ gcs:
   # GCS_CUSTOM_STORAGE_CLASS_MAP, allow setup storage class depends on backup name regexp pattern, format nameRegexp > className
   custom_storage_class_map: {}
   debug: false                 # GCS_DEBUG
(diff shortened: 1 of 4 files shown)


aadant:
@Slach do you mind reviewing this ?

Slach:
@minguyen9988 please renew cli snapshot
```
clear && rm -fv ./test/testflows/clickhouse_backup/tests/snapshots/cli.py.cli.snapshot && RUN_TESTS="/clickhouse backup/cli*" ./test/testflows/run.sh
```

Slach (approved):

minguyen9988:
ok I will do it before the next PR

PR #788, opened 2023-11-17
Fix docs for K8s Job examples - Fix reference to the previous backup

nadberezny:
Current bash script provided in `Examples.md` has an incorrect behaviour when it comes to looking up for the previous backups in `system.backup` table.

The following clause is needed in order to look for backups for a specific shard only:
```AND name LIKE '%${SERVER}%'```


Slach:
nice catch, thanks for contributing

PR #804, opened 2023-12-29
Change watch start behavior for respect last backup time and type

rumanzo:
I make my backups with systemd service and timer using watch functionality.
There was big surprise, that any service (clickhouse-backup) restart by any reason triggers new full backup, that in my case continues ~20 hours and lead to higher backup storage consumption.
I changed this behavior by adding timeout on start and define next backup type full or increment.

### DIFF
--- a/pkg/backup/watch.go
+++ b/pkg/backup/watch.go
@@ -115,7 +115,11 @@ func (b *Backuper) Watch(watchInterval, fullInterval, watchBackupNameTemplate, t
 		b.log.Infof("Time before do full backup %v", timeBeforeDoFullBackup)
 		if timeBeforeDoBackup > 0 && timeBeforeDoFullBackup > 0 {
 			b.log.Infof("Waiting %d seconds until continue doing backups due watch interval", timeBeforeDoBackup)
-			time.Sleep(b.cfg.General.WatchDuration - now.Sub(lastBackup))
+			select {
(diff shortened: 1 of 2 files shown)


Slach:
thanks for contribution

Slach:
could you fix integration tests failures ? https://github.com/Altinity/clickhouse-backup/actions/runs/7355512267/job/20024306085#step:7:2166

Slach:
create `./test/integration/.env` from `.env.example`

you need `docker-compose` for properly test running

use 
```bash
RUN_TESTS=TestServerAPI ./test/integration/run.sh
```
for local development check

rumanzo:
> could you fix integration tests failures ? https://github.com/Altinity/clickhouse-backup/actions/runs/7355512267/job/20024306085#step:7:2166

I made local tests that passed fine. But I'll check these tests and will fix it

rumanzo:
I changed waiting next backup time behavior for respect ctx.Done instead of usual sleep.
Notice that with new logic I changed test expect value too, because new backup waiting watch timeout before continue. It means that if we run watch several times in short time, there will be less completed backups due watch timeout defined in config.
TestServerAPI test passed, also I checked server functionality manually too. All fine, even cancelation in watch wait time for start do first backup.

Please review code logic, code style, etc.

PR #811, opened 2024-01-15
docs: remove suffix -race from usage documentation

sanadhis:
Remove this suffix `-race` from usage doc as the binary apparently doesn't have this `-race` suffix.

Test from docker image **altinity/clickhouse-backup:2.4.17**:
```
ca0659cd2951:/# ls -lah /bin | grep clickhouse
-rwxr-xr-x    1 root     root       46.8M Jan  9 16:13 clickhouse-backup
ca0659cd2951:/# which clickhouse
```

Note: this is my first time opening PR to this repository, so appreciate any feedback!

### DIFF
--- a/Manual.md
+++ b/Manual.md
@@ -1,7 +1,7 @@
 ### CLI command - tables
 ```
 NAME:
-   clickhouse-backup-race tables - List of tables, exclude skip_tables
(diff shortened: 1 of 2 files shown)


Slach:
Nice catch, thank you so much for contribution

PR #893, opened 2024-04-12
docs: removing typo

gmelillo:
Just a typo I've noticed while reading the docs.


Slach:
nice catch, thanks for contribution

PR #922, opened 2024-05-13
Allow stopping api server when watch command is stopped

### LINKED ISSUE #923: API server should restart if `watch` command fails
Hi, with the current `--watch` implementation on the API server, if error occurs in the watch goroutine, process just stops leaving API server running without performing any work. You need to manually restart the server or start watch command by the API call in order to have it running again.
Also `watch` states are not clearly visible, for example if you run backup server in kubernetes, pod will be treated as healthy even if it does not perform expected functionality.

I think it would make sense to be able to configure API server that it is considered healthy only when watch goroutine is running and stop the API server when watch command fails. 
I made a pull request to address this [issue](https://github.com/Altinity/clickhouse-backup/pull/922)

tadus21:
Hi, with the current `--watch` implementation on the API server, if error occurs in the `watch` goroutine, process just stops leaving API server running without performing any work and you need to manually restart the server or `start` watch command by the API call in order to have it running again. 

I think it would make sense to stop API server if `watch` command failed so I am proposing a change that allows running `watch` command as a main API process and once its stopped, whole server is stopped as well. 

Added this feature under a new `watch_is_main_process` config argument so it is backwards compatible and will not introduce new behaviour in the tool without explicitly enabling it.

fix https://github.com/Altinity/clickhouse-backup/issues/923

Does this change make sense and I could expect it to be introduced in the system?


Slach:
build failed
could check `make clean build/linux/amd64/clickhouse-backup` locally?

pkg/config/config.go
```diff
@@ -15,7 +15,6 @@ import (
 	"github.com/apex/log"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/urfave/cli"
-	"gopkg.in/yaml.v3"
```
Slach: why you remove this line?

Slach (changes requested):

pkg/config/config.go
tadus21: oh sorry, my ide for some reason decided that this import is not used, restored it back.

Slach:
for tesflows pass
```
rm -rf ./test/testflows/clickhouse_backup/tests/snapshots/cli.py.cli.snapshot
cp ./test/testflows/.env.example ./test/testflows/.env
```
edit `./test/testflows/.env`

run tests
```
bash -xe ./test/testflows/run.sh
```

commit changed cli.py.cli.snapshot

Slach (approved):

Slach:
All looks good now, thanks for contribution

small step before merge
could you run pull latest master? i'm upgraded github action build.yaml to allow Cleanup step works properly

tadus21:
> All looks good now, thanks for contribution
> 
> small step before merge could you run pull latest master? i'm upgraded github action build.yaml to allow Cleanup step works properly

Thanks for helping with this. I think because I rebased master I need to get a new workflow approval in order to merge this

PR #929, opened 2024-05-30
exit the application once watch is stopped

tadus21:
Fix for [PR](https://github.com/Altinity/clickhouse-backup/pull/922) that I've made couple weeks before. Noticed that my solution actually does not end `Run` function and the process keeps hanging indefinitely. To fix this added another channel which would notify that the server/process needs to be stopped.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -45,6 +45,7 @@ type APIServer struct {
 	config                  *config.Config
 	server                  *http.Server
 	restart                 chan struct{}
+	stop                    chan struct{}
 	metrics                 *metrics.APIMetrics
 	log                     *apexLog.Entry
 	routes                  []string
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -991,11 +998,7 @@ func (api *APIServer) httpWatchHandler(w http.ResponseWriter, r *http.Request) {
 	go func() {
 		b := backup.NewBackuper(cfg)
 		err := b.Watch(watchInterval, fullInterval, watchBackupNameTemplate, tablePattern, partitionsToBackup, schemaOnly, rbacOnly, configsOnly, skipCheckPartsColumns, api.clickhouseBackupVersion, commandId, api.GetMetrics(), api.cliCtx)
-		defer status.Current.Stop(commandId, err)
-		if err != nil {
-			api.log.Errorf("Watch error: %v", err)
-			return
-		}
+		api.handleWatchResponse(commandId, err)
```
tadus21, line 1001: missed this case in the previous PR so fixing it as well

tadus21:
@Slach don't have permissions to add you as a reviewer so tagging you in the comment

Slach (approved):

PR #964, opened 2024-07-29
fix: fix custom List command print invalid zero information

### LINKED ISSUE #963: custom List show zero result
when using rsync to upload/list/download backups, if remote is empty, custom list command get empty, then the final output shows:
```shell
   0B   01/01/0001 00:00:00   remote
````

look into this code `pkg/custom/list_custom.go`, when `strings.TrimRight(out, "\n")` got empty string, `strings.Split("", "\n")` got `[]string{""}` , but we expect it as `[]string{}`.
```shell
outLines := strings.Split(strings.TrimRight(out, "\n"), "\n")
backupList := make([]storage.Backup, len(outLines))
for i, line := range outLines {
	if len(line) > 0 {
		if err = json.Unmarshal([]byte(line), &backupList[i]); err != nil {
			return nil, fmt.Errorf("JSON parsing '%s' error: %v ", line, err)
		}
	}
}
```

and this result to `backupList` got `[]storage.Backup{ storage.Backup{} }`, which contains only one empty struct item. and  lead to the problem above.

straysh:
close https://github.com/Altinity/clickhouse-backup/issues/963

### DIFF
--- a/pkg/custom/list_custom.go
+++ b/pkg/custom/list_custom.go
@@ -23,7 +23,12 @@ func List(ctx context.Context, cfg *config.Config) ([]storage.Backup, error) {
 	args := ApplyCommandTemplate(cfg.Custom.ListCommand, templateData)
 	out, err := utils.ExecCmdOut(ctx, cfg.Custom.CommandTimeoutDuration, args[0], args[1:]...)
 	if err == nil {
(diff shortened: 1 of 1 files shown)


Slach:
thanks for contribution

PR #1000, opened 2024-08-29
docs: update clickhouse-backup docker args example

christopherdeutsch:
Hi! I think one of the examples can be simplified.

The current command in the example for clickhouse-backup is
```
 command:
    - bash
    - -xc
    - "/bin/clickhouse-backup server"
```

And the Dockerfile ends with:
```
ENTRYPOINT ["/entrypoint.sh"]
CMD [ "/bin/clickhouse-backup", "--help" ]
```

This PR will update the docs to just pass `server` to the entrypoint script, since `bash` is unnecessary. The `-xc` flags could be useful for debugging, but I propose that using the args to specify `server` makes the example more clear.

### DIFF
--- a/Examples.md
+++ b/Examples.md
@@ -182,7 +182,7 @@ spec:
                       - clickhouse-server
                       - --config-file=/etc/clickhouse-server/config.xml
                  - name: clickhouse-backup
-                   image: altinity/clickhouse-backup:master
+                   image: altinity/clickhouse-backup:latest
                    imagePullPolicy: Always
                    args: ["server"]
                    env:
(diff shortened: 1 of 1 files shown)


Examples.md
```diff
@@ -184,10 +184,7 @@ spec:
                  - name: clickhouse-backup
                    image: altinity/clickhouse-backup:master
```
Slach: Could you change `:master` to `:latest` in whole Examples.md?

Slach (changes requested):

Examples.md
christopherdeutsch: Done! 👍

Slach:
thanks for contribution

PR #1009, opened 2024-09-12
Corrected log level warn -> warning

NikLoginov:
Fixed log level for warning


Slach:
Good catch, thanks a lot for contribution.

PR #1081, opened 2025-01-30
Pgalbraith backup doc container doc fix

CaptTofu:
I needed to use `clickhouse-server` container image instead of `clickhouse-client` image, which the latter didn't work in aws-dev-eu because of it being arm64. I also elaborated how the process works a bit more.


Slach:
looks weird, how did you make 24 commits instead of just last two

Examples.md
```diff
@@ -319,7 +321,7 @@ spec:
           restartPolicy: Never
           containers:
             - name: run-backup-cron
-              image: clickhouse/clickhouse-client:latest
+              image: clickhouse/clickhouse-server:latest
```
Slach, line 324: let's add comment here about clickhouse-client doesn't support anymore, to avoid comments about "ah why you run second clickhouse-server?"

actually old clickhouse-clients should connect to latest clickhouse-server, protocol compatible, which kind of issue did you look here?

Slach (changes requested):

CaptTofu:
> looks weird, how did you make 24 commits instead of just last two

yes, I didn't realize this. I did a rebase of my fork, I'll redo this.

Examples.md
CaptTofu: my issue was that the container in the cron was failing with an odd 'bash exec' issue when I tried running the container in kubernetes with a simple bash while loop to test in aws-dev-eu (which the nodes are arm64). Some searching and a test on my own laptop (which is arm64) gave me the same error, then I realized that the `clickhouse-client` container didn't run on arm64. Both Anselmo and Josh pointed out what I had suspected and suggested clickhouse-server

Slach: ok. got it
https://hub.docker.com/r/clickhouse/clickhouse-client/tags
doesn't support arm64, lets add comment here

PR #1082, opened 2025-01-30
use clickhouse-server container image instead of clickhouse-client

CaptTofu:
1. clickhouse-client container for the cron job wouldn't run in aws-dev-eu, which uses arm64, switched to clickhouse-server container and it works
2. Added more information and elaboration for the backup process using the operator

### DIFF
--- a/Examples.md
+++ b/Examples.md
@@ -132,7 +132,9 @@ You can use clickhouse-backup for creating periodical backups and keep it local.
 In addition, you may create instance of ClickHouse on another DC and have it fresh by clickhouse-copier to protect you from hardware or DC failures.
 
 ## How to use clickhouse-backup in Kubernetes
-Install the [clickhouse kubernetes operator](https://github.com/Altinity/clickhouse-operator/) and use the following manifest:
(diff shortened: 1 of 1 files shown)


CaptTofu:
Re-created this PR because when I rebased my fork with master, somehow it pulled in recent changes which showed up in the original PR.

Examples.md
```diff
@@ -319,7 +323,7 @@ spec:
           restartPolicy: Never
           containers:
             - name: run-backup-cron
-              image: clickhouse/clickhouse-client:latest
+              image: clickhouse/clickhouse-server:latest
```
Slach, line 326: old `clickhouse-client` should works with new `clickhouse-server`
which kind of issue you got here?

Slach (changes requested):

Examples.md
Slach: ok. got it
https://hub.docker.com/r/clickhouse/clickhouse-client/tags
doesn't support arm64
let's add comment here

Slach (approved):

Slach:
thanks for contribution

PR #1130, opened 2025-04-10
Fix RBAC Backup Query: Change Type Filter to support "local directory" and "local_directory"

yudhiesh:
# What's Changed in this PR?
This PR addresses an issue with RBAC backups where user accounts weren't being restored properly. I noticed that backups failed to restore users stored under the `local_directory` entry in the ClickHouse `system.user_directories` table. The root cause was that the code was filtering on the incorrect type value `local directory` (with a space) instead of the actual value `local_directory` (with an underscore). On my cluster when I query `SELECT * from system.user_directories` it returns the following:
Screenshot 2025-04-10 at 15 30 47
This mismatch caused the query to return no results, leading to backup failures on certain filesystems such as ZFS.
## Changes
- Updated the SQL query in the backup code to filter on `type='local_directory'` instead of `type='local directory'`.
- This ensures the `JSONExtractString` function correctly extracts the access path from the `params` JSON field, returning the proper value (e.g., `/mnt/zfs0/clickhouse/access/`).

### DIFF
--- a/pkg/clickhouse/clickhouse.go
+++ b/pkg/clickhouse/clickhouse.go
@@ -1139,7 +1139,7 @@ func (ch *ClickHouse) GetAccessManagementPath(ctx context.Context, disks []Disk)
 	rows := make([]struct {
 		AccessPath string `ch:"access_path"`
 	}, 0)
-	if err := ch.SelectContext(ctx, &rows, "SELECT JSONExtractString(params,'path') AS access_path FROM system.user_directories WHERE type='local directory'"); err != nil || len(rows) == 0 {
+	if err := ch.SelectContext(ctx, &rows, "SELECT JSONExtractString(params,'path') AS access_path FROM system.user_directories WHERE type='local_directory'"); err != nil || len(rows) == 0 {
 		configFile, doc, err := ch.ParseXML(ctx, "config.xml")
 		if err != nil {
 			log.Warn().Msgf("can't parse config.xml from %s, error: %v", configFile, err)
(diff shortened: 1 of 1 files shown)


pkg/clickhouse/clickhouse.go
```diff
@@ -1139,7 +1139,7 @@ func (ch *ClickHouse) GetAccessManagementPath(ctx context.Context, disks []Disk)
 	rows := make([]struct {
 		AccessPath string `ch:"access_path"`
 	}, 0)
-	if err := ch.SelectContext(ctx, &rows, "SELECT JSONExtractString(params,'path') AS access_path FROM system.user_directories WHERE type='local directory'"); err != nil || len(rows) == 0 {
+	if err := ch.SelectContext(ctx, &rows, "SELECT JSONExtractString(params,'path') AS access_path FROM system.user_directories WHERE type='local_directory'"); err != nil || len(rows) == 0 {
```
Slach: looks like different versions of clickhouse have different value for `type` field
which version do you use?

better use `type IN ('local_directory','local directory')`

Slach (changes requested):

pkg/clickhouse/clickhouse.go
yudhiesh: Ah makes sense, I am using a much older version(`23.x`) but will be upgrading soon although it will take time to verify things work as usual + ensure all workflows on cluster work due to backwards incompatibility. So I hope that we can add this change to support backups.

Slach: https://fiddle.clickhouse.com/db2ed387-b7c1-4453-9d42-19f7796a2428

yudhiesh: This is the version that I am using https://fiddle.clickhouse.com/76d75b26-c3b9-42aa-8b5f-64ff944df031 even on the latest version it follows `local_directory` https://fiddle.clickhouse.com/b870912f-325a-46ff-9566-ddfa071e0a63

Slach (approved):

PR #1159, opened 2025-06-02
Checking that the local backup hasn't already been deleted by the cre…

### LINKED ISSUE #1157: `server --watch` aborts watch process when `backups_to_keep_local: -1` and `--watch-delete-source=false`
When running `clickhouse-backup server --watch` with `backups_to_keep_local: -1` there are duplicate attempts to delete the local backup and the second one fails due to the first one successfully removing the local backup.

Logs:
```console
2025-05-30 02:22:40.671 INF pkg/backup/upload.go:256 > done backup=2025053002-shard0-full duration=14.038s object_disk_size=0B operation=upload upload_size=71.15MiB version=2.6.18
2025-05-30 02:22:40.869 INF pkg/storage/general.go:163 > list_duration=180.019109
2025-05-30 02:22:40.869 INF pkg/backup/delete.go:157 > remove '/var/lib/clickhouse/backup/2025053002-shard0-full'
2025-05-30 02:22:41.013 INF pkg/backup/delete.go:166 > done backup=2025053002-shard0-full duration=334ms location=local operation=delete
2025-05-30 02:22:41.023 ERR pkg/server/metrics/metrics.go:246 > metrics.ExecuteWithMetrics(delete) return error: '2025053002-shard0-full' is not found on local storage
2025-05-30 02:22:41.024 ERR pkg/server/server.go:618 > Watch error: too many errors create_remote: 0, delete local: 1, during watch full_interval: 168h, abort watching
```

The first delete is successful [here](https://github.com/Altinity/clickhouse-backup/blob/master/pkg/backup/upload.go#L263) as part of the upload process but then it tries to delete it again [here](https://github.com/Altinity/clickhouse-backup/blob/master/pkg/backup/watch.go#L127) which fails

I think you could alter the first delete to:
```go
if err = b.RemoveOldBackupsLocal(ctx, !deleteSource, nil); err != nil {
   return fmt.Errorf("can't remove old local backups: %v", err)
}
```

This would keep the most recent backup from the during the upload step which would then be deleted in the `Watch` function. Alternatively, you could stop the delete backup from throwing an error if the backup doesn't exist locally

elliott-logrocket:
If backups_to_keep_local=-1 then RemoveOldBackupsLocal will remove any local backups RemoveOldBackupsLocal is called on every call to Upload (so whenever CreateToRemote is called, so is RemoveOldBackupsLocal) This causes an issue when --watch-delete-source=false since Watch will also try to delete the new local backup after the CreateToRemote call finishes By adding a check that BackupsToKeepLocal is greater than zero we make sure that we do not get duplicate deletes which error out and cause Watch to abort erroneously

Closes #1157

### DIFF
--- a/pkg/backup/watch.go
+++ b/pkg/backup/watch.go
@@ -124,7 +124,8 @@ func (b *Backuper) Watch(watchInterval, fullInterval, watchBackupNameTemplate, t
 				createRemoteErr, createRemoteErrCount = metrics.ExecuteWithMetrics("create_remote", createRemoteErrCount, func() error {
 					return b.CreateToRemote(backupName, deleteSource, "", diffFromRemote, tablePattern, partitions, skipProjections, schemaOnly, backupRBAC, false, backupConfigs, false, skipCheckPartsColumns, false, version, commandId)
 				})
-				if !deleteSource {
+				// If backups_to_keep_local=-1 then the local backup is deleted in the upload step when RemoveOldBackupsLocal is called
+				if !deleteSource && b.cfg.General.BackupsToKeepLocal >= 0 {
 					deleteLocalErr, deleteLocalErrCount = metrics.ExecuteWithMetrics("delete", deleteLocalErrCount, func() error {
 						return b.RemoveBackupLocal(ctx, backupName, nil)
 					})


Slach:
thanks for contribution

PR #1246, opened 2025-09-17
feat: add GCS ServiceAccount support

kamushadenes:
This PR adds support for using GCS ServiceAccounts instead of keys, when the default token source has the permission to impersonate the SA.

This is meant to be used inside Compute instances that have the necessary permission or inside Kubernetes clusters where the K8S ServiceAccount is linked to a ServiceAccount that can do the impersonation.


Slach (approved):

Slach:
fixed https://github.com/Altinity/clickhouse-backup/issues/1173

PR #1247, opened 2025-09-18
S3 chunksize

KimDoKy:
Background
We are backing up ClickHouse databases to MINIo object storage in a private cloud environment.

Problem
When using the existing max_parts_count configuration, we encountered excessive load during backup operations. This caused resource contention issues and slowed down uploads, especially with larger backups.

Solution
I added a new option to control chunk size during object storage uploads. By adjusting the chunk size, we can optimize resource usage and avoid spikes in network or disk load. This feature has been tested successfully in our environment.

Update
When the file size was divided based on chunk_size, it was modified not to exceed 10000.

Request
I hope this chunk size control feature can be officially included in the project to provide better flexibility for MINIo and similar scenarios.

Please let me know if any changes are needed or if further testing is required. Thank you!


KimDoKy:
An error was found.

PR #1266, opened 2025-10-10
remove s3 debug from example and set log level to info

weskerfoot:
This removes the setting in the k8s example config for setting the S3 log level to debug, and changes the LOG_LEVEL to info.

I found that if you actually deploy it with the settings in the example the S3 logging uses up quite a lot of resources and can even cause the system load average to go up on a live cluster.

### DIFF
--- a/Examples.md
+++ b/Examples.md
@@ -241,7 +241,7 @@ spec:
               args: [ "server" ]
               env:
                 - name: LOG_LEVEL
-                  value: "debug"
+                  value: "info"
                 - name: ALLOW_EMPTY_BACKUPS
                   value: "true"
                 - name: API_LISTEN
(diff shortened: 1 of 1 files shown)


Examples.md
```diff
@@ -274,8 +274,6 @@ spec:
                 # remove it for production S3
                 - name: S3_DISABLE_SSL
                   value: "true"
-                - name: S3_DEBUG
-                  value: "true"
```
Slach, line 278: let's stay it but change to "false"
when somebody will read it, it allow to understand about available options

cause nobody read README ;(

Slach (changes requested):

Examples.md
weskerfoot: sure, that makes sense, done

PR #1310, opened 2025-12-20
fix S3 multipart operations to respect S3_MAX_PARTS_COUNT

xavierleune:
Hi there,

I'm using a s3 provider with a limit of 1000 parts per upload, unfortunately the operations use the hardcoded limit for AWS (10000) instead of using the configValue S3_MAX_PARTS_COUNT.
This pull request should fix that behavior.

Thanks,


Slach (approved):
thanks for contribution, let's wait when CI jobs finish

xavierleune:
Hello @Slach, thanks for the merge ! Any chance we might get a release soon ? Thanks !

PR #1312, opened 2025-12-24
feat: add TLS support for Keeper connections

### LINKED ISSUE #1311: support tls for keeper and clickhouse communication
To support RBAC backups in environments where ClickHouse Keeper or ZooKeeper enforces TLS, the communication between ClickHouse and Keeper may need to support TLS.

loomts:
Resolved #1311 , to simplify the config, I just reuse the tls config of clickhouse.
According to https://clickhouse.com/docs/operations/ssl-zookeeper, the sample use OS RootCAs, which can also be handle.

### DIFF
--- a/pkg/backup/create.go
+++ b/pkg/backup/create.go
@@ -794,7 +794,7 @@ func (b *Backuper) createBackupNamedCollections(ctx context.Context, backupPath 
 			}
 
 			k := keeper.Keeper{}
-			if err = k.Connect(ctx, b.ch); err != nil {
+			if err = k.Connect(ctx, b.ch, b.cfg); err != nil {
 				return 0, err
 			}
 			defer k.Close()
@@ -844,7 +844,7 @@ func (b *Backuper) createBackupRBACReplicated(ctx context.Context, rbacBackup st
 	rbacDataSize := uint64(0)
 	if err = b.ch.SelectContext(ctx, &replicatedRBAC, "SELECT name FROM system.user_directories WHERE type='replicated'"); err == nil && len(replicatedRBAC) > 0 {
 		k := keeper.Keeper{}
-		if err = k.Connect(ctx, b.ch); err != nil {
+		if err = k.Connect(ctx, b.ch, b.cfg); err != nil {
 			return 0, err
 		}
 		defer k.Close()
--- a/pkg/backup/create.go
+++ b/pkg/backup/create.go
@@ -794,7 +794,7 @@ func (b *Backuper) createBackupNamedCollections(ctx context.Context, backupPath 
 			}
 
 			k := keeper.Keeper{}
-			if err = k.Connect(ctx, b.ch, b.cfg); err != nil {
+			if err = k.Connect(ctx, b.ch); err != nil {
 				return 0, err
 			}
 			defer k.Close()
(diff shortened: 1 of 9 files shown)


pkg/keeper/keeper.go
```diff
@@ -93,11 +98,37 @@ func (k *Keeper) Connect(ctx context.Context, ch *clickhouse.ClickHouse) error {
...
+		}
 		keeperHosts[i] = fmt.Sprintf("%s:%s", hostNode.InnerText(), port)
 	}
-	conn, _, err := zk.Connect(keeperHosts, sessionTimeout, zk.WithLogger(newKeeperLogger()))
-	if err != nil {
-		return err
+	var conn *zk.Conn
+	if isSecure {
+		log.Info().Msgf("isSecure=%v, keeperHosts=%v, caPath=%v, certPath=%v, keyPath=%v, skipVerify=%v, use TLS for keeper connection", isSecure, keeperHosts, cfg.ClickHouse.TLSCa, cfg.ClickHouse.TLSCert, cfg.ClickHouse.TLSKey, cfg.ClickHouse.SkipVerify)
+		tlsConfig, err := utils.NewTLSConfig(cfg.ClickHouse.TLSCa, cfg.ClickHouse.TLSCert, cfg.ClickHouse.TLSKey, cfg.ClickHouse.SkipVerify)
```
Slach: Not sure that this is a good idea to use the same certificates for connecting `clickHouse-backup -> clickhouse` and `clickhouse-backup -> keeper`,

according to https://clickhouse.com/docs/operations/ssl-zookeeper

better parse `/*/client/*` from `/var/lib/clickhouse/preprocessed_configs/config.xml`
when isSecure true

Slach (changes requested):
please add parsing `/*/client/*` from `/var/lib/clickhouse/preprocessed_configs/config.xml`
when `isSecure` true instead of using clickhouse TLS connection setting

moreover, could you add `zookeeper` and `keeper` tls settings to `/tests/integration/docker-compose*.yml` and `/tests/integration/dynamic_settings.sh`


and test it via
`RUN_TESTS=^TestS3$ ./test/integration/run.sh`

Slach:
thanks you for your contribution

pkg/keeper/keeper.go
loomts: got it, I will fix it soon, thank you for you guidance.

loomts: got it, I will fix it soon, thank you for you guidance.

Slach:
@loomts any news about requested changes?

pkg/backup/create.go
```diff
@@ -807,7 +794,7 @@ func (b *Backuper) createBackupNamedCollections(ctx context.Context, backupPath
 			}
 
 			k := keeper.Keeper{}
-			if err = k.Connect(ctx, b.ch); err != nil {
+			if err = k.Connect(ctx, b.ch, b.cfg); err != nil {
```
Slach: i think this is unnecessary cause we can parse config in k.Connect

Slach (changes requested):

pkg/backup/create.go
loomts: fixed

loomts:
> @loomts any news about requested changes?

@Slach I fixed according to the comments, PTAL.

Slach (approved):
thanks a lot for your contribution, looks great!

PR #1330, opened 2026-03-04
fix(s3): fallback retry with Content-MD5 for DeleteObjects

### LINKED ISSUE #1329: Regression in 2.6.42: S3 DeleteObjects fails on Huawei S3 (missing Content-MD5)
## Problem
After upgrading to `2.6.42`, deleting remote backups on some S3-compatible providers fails because the new batch delete path (`DeleteObjects`) is used by default.

Example error:

```
api error InvalidRequest: Missing required header for this request: Content-MD5
```

This is reproducible on Huawei S3.

## Why this regressed in 2.6.42
`2.6.42` introduced batch deletion via `DeleteObjects`. Some S3-compatible providers require `Content-MD5` for this API call. Previous versions (`2.6.41` and earlier) used one-by-one `DeleteObject` and did not hit this requirement.

## Current workaround
Set:

```yaml
s3:
  request_content_md5: true
```

## Expected behavior
When batch delete fails specifically due to missing `Content-MD5`, clickhouse-backup should retry once with `Content-MD5` automatically, so existing installations keep working after upgrade.

## Acceptance criteria
- `RemoveBackupRemote` with S3 batch delete succeeds on providers that require `Content-MD5`.
- No behavior change for AWS S3 and providers that do not require it.
- Existing explicit `request_content_md5: true` still works.

BorisTyshkevich:
## Summary
- fix regression introduced by batch `DeleteObjects` in 2.6.42 for S3-compatible providers requiring `Content-MD5` (e.g. Huawei S3)
- keep existing `request_content_md5: true` behavior unchanged
- add helper + tests for detecting missing Content-MD5 response

## Details
When `DeleteObjects` fails with `InvalidRequest` + missing `Content-MD5`, retry the same batch once with MD5 middleware enabled.

## Validation
- `go test ./pkg/storage -run TestIsDeleteObjectsMissingContentMD5Error -v`
- `go test ./pkg/storage`

Closes #1329


Slach:
did you see https://github.com/Altinity/clickhouse-backup/commit/d18eef77c78add5bf43822ba1b51125824fb02d9 ?

BorisTyshkevich:
yes, and already recommended to the client

Slach:
ok. will merge master and will merge later

PR #1335, opened 2026-03-24
fix(s3): wrap credentials with NewCredentialsCache

kakao-tars-case:
## Summary
- Wrap AWS credentials with `aws.NewCredentialsCache()` to cache resolved credentials and avoid redundant credential resolution on every API call
- Also cache credentials before passing to STS `AssumeRoleProvider` in IRSA + AssumeRole flow

## Details
Without `NewCredentialsCache`, each S3 API call may trigger a fresh credential resolution (e.g., hitting IMDS or STS endpoint). This adds latency and can cause throttling under high request volume.

## Validation
- `go test ./pkg/storage/ -v` — all unit tests pass
- `go build ./...` — build succeeds


Slach:
thanks for contibution

Slach (approved):

PR #1351, opened 2026-04-08
feat: add force_rebalance option to distribute data across JBOD disks

### LINKED ISSUE #1350: Disk rebalancing does not trigger when backup source disk name exists on target (e.g. "default")
## Problem

When restoring a backup taken on a single-disk machine (all data on the `default` disk) to a target machine with multiple JBOD disks under the same `default` storage policy, the disk rebalancing logic in `reBalanceTablesMetadataIfDiskNotExists()` never triggers.

### Source (backup) machine
- Single large volume
- All data stored on the built-in `default` disk at `/var/lib/clickhouse/`
- Tables use `storage_policy = 'default'` (or no explicit policy)

### Target (restore) machine
- 4 smaller JBOD disks configured as:
```xml
<storage_configuration>
    <disks>
        <hyperdisk00><path>/var/lib/private/clickhouse/hyperdisk00/</path><type>local</type></hyperdisk00>
        <hyperdisk01><path>/var/lib/private/clickhouse/hyperdisk01/</path><type>local</type></hyperdisk01>
        <hyperdisk02><path>/var/lib/private/clickhouse/hyperdisk02/</path><type>local</type></hyperdisk02>
        <hyperdisk03><path>/var/lib/private/clickhouse/hyperdisk03/</path><type>local</type></hyperdisk03>
    </disks>
    <policies>
        <default>
            <volumes>
                <default>
                    <disk>hyperdisk00</disk>
                    <disk>hyperdisk01</disk>
                    <disk>hyperdisk02</disk>
                    <disk>hyperdisk03</disk>
                    <load_balancing>least_used</load_balancing>
                </default>
            </volumes>
        </default>
    </policies>
</storage_configuration>
```

## Root cause

The rebalancing check at `download.go:374` is:
```go
if _, diskExists := b.DiskToPathMap[disk]; !diskExists && disk != b.cfg.ClickHouse.EmbeddedBackupDisk {
```

Since the `default` disk **always exists** on every ClickHouse instance (it's the built-in disk at `/var/lib/clickhouse/`), `diskExists` is always `true` and the rebalancing block is never entered. All parts are downloaded to the small `default` disk instead of being distributed across the 4 JBOD disks.

## Current workaround

The only workaround is `disk_mapping`, but it can only redirect `default` to a single path — it cannot distribute data across multiple JBOD disks:
```yaml
clickhouse:
  disk_mapping:
    default: /var/lib/private/clickhouse/hyperdisk00/
```

This defeats the purpose of having a JBOD setup.

## Additionally

Even when the rebalancing block is entered (non-existent disk name), two places in `downloadTableData()` and `findDiffFileExist()` only check `RebalancedDisk` when the original disk doesn't exist. If the original disk exists but rebalancing was forced, the `RebalancedDisk` value is ignored and data still goes to the wrong disk.

## Proposed fix

A new config option `force_rebalance: true` (env: `CLICKHOUSE_FORCE_REBALANCE`) that:

1. Forces entry into the rebalancing block even when the disk name exists
2. Prioritizes `RebalancedDisk` over the original disk name everywhere in the download path

This allows the existing `least_used` strategy to distribute parts across all disks in the target's storage policy. See attached PR.

SLoeuillet:
## Summary

Fixes #1350

When restoring a backup taken on a single-disk machine to a target with multiple JBOD disks under the same `default` storage policy, the disk rebalancing logic never triggers because the `default` disk name exists on both machines. All data ends up on the small built-in `default` disk instead of being distributed across JBOD disks.

The existing `disk_mapping` workaround can only redirect to a **single** path — it cannot distribute across multiple disks.

### New config option: `force_rebalance`

**Config:** `clickhouse.force_rebalance: true` (env: `CLICKHOUSE_FORCE_REBALANCE`)

Forces the rebalancing code path during `download` even when the backup's disk name exists on the target, allowing distribution across JBOD disks under the same storage policy using the existing `least_used` strategy.

### Download path fixes

- **`reBalanceTablesMetadataIfDiskNotExists()`**: enter the rebalancing block when `force_rebalance: true`, even if the disk exists in `DiskToPathMap`
- **`downloadTableData()` (directory format)**: when `RebalancedDisk` is set, always use it to determine the local download path and recompute `tableLocalPath`
- **`findDiffFileExist()`**: prioritize `RebalancedDisk` when set for incremental diff files
- **Resumable downloads**: skip rebalancing for parts that already have a `RebalancedDisk` from a previous run, preventing metadata/file mismatch when `FreeSpace` changes between resume attempts

### Restore path fixes

- **`HardlinkBackupPartsToStorage()`**: use a per-iteration `activeDisk` variable instead of mutating `backupDiskName` across inner loop iterations (the old code corrupted disk assignments for subsequent parts)
- **`HardlinkBackupPartsToStorage()`**: when the rebalanced disk is not in `dstDataPaths`, build the `store/{uuid}/detached/` path directly from the disk path and table UUID. ClickHouse creates `store/{uuid}/detached/` on **all** disks of the storage policy at table load time ([MergeTreeData.cpp:443-444](https://github.com/ClickHouse/ClickHouse/blob/master/src/Storages/MergeTree/MergeTreeData.cpp)), so `ATTACH PART` via `tryGetDiskForDetachedPart()` will find parts on any policy disk. This keeps hardlinks on the same filesystem (no cross-device copy) and preserves the JBOD distribution from download.

### Documentation & test

- Document `force_rebalance` in README config explained section
- Add `TestForceRebalance` e2e integration test

## Use case

**Source** (backup): single large volume, everything on `default` disk

**Target** (restore): 4 JBOD disks under the `default` storage policy
```xml
<disks>
    <hyperdisk00><path>/var/lib/private/clickhouse/hyperdisk00/</path></hyperdisk00>
    <hyperdisk01><path>/var/lib/private/clickhouse/hyperdisk01/</path></hyperdisk01>
    <hyperdisk02><path>/var/lib/private/clickhouse/hyperdisk02/</path></hyperdisk02>
    <hyperdisk03><path>/var/lib/private/clickhouse/hyperdisk03/</path></hyperdisk03>
</disks>
<policies>
    <default>
        <volumes><default>
            <disk>hyperdisk00</disk><disk>hyperdisk01</disk>
            <disk>hyperdisk02</disk><disk>hyperdisk03</disk>
            <load_balancing>least_used</load_balancing>
        </default></volumes>
    </default>
</policies>
```

**Config:**
```yaml
clickhouse:
  force_rebalance: true
```

### How it works

1. **Download** distributes parts across the 4 hyperdisks using `least_used` strategy
2. **Restore** hardlinks each part from its backup location into `store/{uuid}/detached/` on the **same disk** (same filesystem, no cross-device copy)
3. **`ATTACH PART`** scans all disks of the storage policy ([tryGetDiskForDetachedPart](https://github.com/ClickHouse/ClickHouse/blob/master/src/Storages/MergeTree/MergeTreeData.cpp)), finds each part on its respective disk
4. Data stays distributed across the JBOD — no post-restore redistribution needed

## Test plan

- [ ] Verify build passes (`go build ./cmd/clickhouse-backup/`)
- [ ] Test download+restore from single-disk backup to multi-disk JBOD target with `force_rebalance: true`
- [ ] Verify parts are distributed across all JBOD disks after restore
- [ ] Verify behavior is unchanged when `force_rebalance` is not set (default: `false`)
- [ ] Test with `--resumable` downloads (interrupt + resume should preserve disk assignments)
- [ ] Test with incremental backups

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Slach (changes requested):
please provide separate e2e test in ./test/integration/*_test.go

pkg/config/config.go
```diff
@@ -274,6 +274,9 @@ type ClickHouseConfig struct {
 	TLSCa                            string            `yaml:"tls_ca" envconfig:"CLICKHOUSE_TLS_CA"`
 	MaxConnections                   int               `yaml:"max_connections" envconfig:"CLICKHOUSE_MAX_CONNECTIONS"`
 	Debug                            bool              `yaml:"debug" envconfig:"CLICKHOUSE_DEBUG"`
+	// ForceRebalance triggers disk rebalancing during download even when the backup's disk
+	// name exists on the target, allowing distribution across JBOD disks under the same storage policy
+	ForceRebalance                   bool              `yaml:"force_rebalance" envconfig:"CLICKHOUSE_FORCE_REBALANCE"`
```
Slach, line 279: also add description to REAMDE.md in config explained section

Slach (changes requested):

Slach:
I updated test/testflows/clickhouse_backup/tests/snapshots/cli.py.cli.snapshot after you changed configuaration

SLoeuillet:
## CI status update

Fixed `TestForceRebalance` failures on older ClickHouse versions (20.3, 20.8):
- `toDateTime('2024-01-01')` → `toDateTime('2024-01-01 00:00:00')` — CH < 22.x can't parse date-only strings
- `<clickhouse>` → `<yandex>` XML root tag — for compatibility with all versions (still supported on 26.3)

The `TestRestoreMutationInProgress` failure on 22.3 is **unrelated** to this PR — it's a pre-existing flaky test (`Cannot parse string 'a' as UInt64` during mutation restore).

SLoeuillet:
## CI run #2 update

Fixed `TestForceRebalance` failure on 20.3/20.8: `DROP TABLE ... SYNC` is not supported before CH 21.x. Now conditional.

Remaining failures are **unrelated** to this PR:
- **21.3**: `TestRestoreDistributedCluster` (188s timeout)
- **25.8**: `TestRBAC`

`TestForceRebalance` itself **passes** on 21.3 (78s) and 25.8 (128s).

Slach:
@SLoeuillet thanks for your efforts, yep confirm some test could failed on CI/CD

use 
`LOG_LEVEL=XXX CLICKHOUSE_VERSION=XXX RUN_TESTS=TestRegExp ./test/integration/run.sh`
for local running

SLoeuillet:
## Status: on hold

The `force_rebalance` feature works correctly for **download** — parts are distributed across JBOD disks and metadata is consistent (including with `--resumable`).

However, **restore fails** with a deeper issue:

### Root cause

`HardlinkBackupPartsToStorage` receives `dstTable.DataPaths` from `system.tables.data_paths`, which only contains the disk(s) where ClickHouse has actually placed data for this table. For a freshly created table (schema-only restore), ClickHouse assigns it to a **single** disk via `least_used`. So `dstDataPaths` only has one entry (e.g. `hyperdisk02`), but the backup has parts rebalanced across all 4 disks.

When the restore tries `dstDataPaths["hyperdisk03"]` for a part that was downloaded to `hyperdisk03`, it either:
- Fails with "not contains rebalanced hyperdisk03"
- Or (with the variable corruption bug pre-fix) falls through to the wrong disk

### What works now
- `clickhouse-backup download` with `force_rebalance: true` correctly distributes parts across JBOD disks
- `--resumable` downloads preserve rebalancing assignments across interruptions

### What needs fixing
- `restore` needs to build `dstDataPaths` from all disks in the table's storage policy, not just from `system.tables.data_paths` which only reflects where ClickHouse has already written data. This likely requires querying `system.storage_policies` + `system.disks` to get all possible target paths for the table's policy.

### Workaround
For now, users can:
1. Download with `force_rebalance: true` (works)
2. Restore without `force_rebalance` — manually move parts or let ClickHouse redistribute after initial restore to a single disk

SLoeuillet:
## Status: ready for review

`TestForceRebalance` passes locally on ClickHouse 24.8 with race detector enabled (82s).

The "on hold" status from earlier comments is lifted — the restore path now correctly hardlinks parts into `store/{uuid}/detached/` on the rebalanced disk (same filesystem), and `ATTACH PART` finds them via `tryGetDiskForDetachedPart()` which scans all disks of the storage policy.

PR description has been updated to reflect the final implementation. History squashed to 2 clean commits.

Slach (approved):

SLoeuillet:
@Slach thanks

Issue #1324, opened 2026-02-03
Missing required header for this request: Content-MD5 when I DeleteObjects from S3


hung2602:
Below is the specific log when I using config backups_to_keep_remote: 3 to rotate backups stored on S3.
2026-02-03 09:33:13.838 WRN pkg/backup/backuper.go:71 > Will wait near 5.532522239s and retry, error=S3 batch delete failed for batch starting at index 0: DeleteObjects API call failed: operation error S3: DeleteObjects, https response error StatusCode: 400, RequestID: ac1044ee:19b8cc8520a:34818:892f, HostID: 55f9344b87d2cba37d410d59e01a2548bae766e7644ffbbfd464d0e5a3892b0f, api error InvalidRequest: Missing required header for this request: Content-MD5
Is there a way to fix this?

hung2602:
I used the S3 compatible.

Slach:
@hung2602 which clickhouse-backup version do you use?

hung2602:
clickhouse-backup --version
Version:         2.6.42
Git Commit:      6aa5680cb78cc2fdb29c5a0a5a2859b487f7c22a
Build Date:      2026-01-29
@Slach I has used clickhouse-backup version 2.6.42

vanyasvl:
@Slach when we can expect 2.6.43 with this fix?

Issue #829, opened 2024-02-12
upload doesn't work for S3 when object lock turned on by default


Slach:
```
InvalidRequest: Content-MD5 OR x-amz-checksum- HTTP header is required for Put Object requests with Object Lock parameters
```

UnamedRus:
https://github.com/aws/aws-sdk-go-v2/issues/1689
https://github.com/aws/aws-sdk-go-v2/pull/2040

Looks like, it just matter of setting value to `ChecksumAlgorithm` parameter

PR #1363, opened 2026-04-22
fix: attach per-part checksum header on UploadPart for Object Lock buckets

david-pivonka:
## Summary

Fixes upload failures against S3 buckets with Object Lock enabled.

## Problem

With `s3.checksum_algorithm: CRC32` set in config, single `PutObject` / `CopyObject` uploads succeed against Object Lock buckets (that path was fixed in #829). Large file uploads still fail during multipart:

```
operation error S3: UploadPart, https response error StatusCode: 400,
api error InvalidRequest: Content-MD5 OR x-amz-checksum-* HTTP header
is required for Put Part requests with Object Lock parameters
```

`s3manager.Uploader` sends the per-part checksum as an HTTP trailer (via aws-chunked encoding). Object Lock rejects trailer-based checksums on `UploadPart` — it requires the checksum as a request header.

## Fix

When `CheckSumAlgorithm == CRC32` and the file is large enough to trigger multipart, `PutFileAbsolute` dispatches to a manual multipart path (`putFileMultipartCRC32`) that:

1. Calls `CreateMultipartUpload` (existing enrich helper used, so tags / SSE / ACL preserved).
2. Reads each part into a buffer, computes CRC32 (IEEE), base64-encodes it.
3. Calls `UploadPart` with `ChecksumAlgorithm: CRC32` and `ChecksumCRC32: <base64>` — the SDK serialises this as the `x-amz-checksum-crc32` request header, which Object Lock accepts.
4. Aborts on any failure; completes via `CompleteMultipartUpload` on success.

For all other cases (no checksum configured, non-CRC32 algo, file below part size), the original `s3manager.Uploader` path is unchanged.

## Scope

- Only activates when the operator opts in via `checksum_algorithm: CRC32`, so existing deployments are unaffected.
- Does not touch `CopyObject` (multipart copy) — `UploadPartCopyInput` has no `ChecksumAlgorithm` field; server-side copy derives integrity from the source object and is not covered by this failure mode.
- Sequential upload loop; concurrency matches `s3manager.Uploader` behaviour only in the dispatched case. Can be parallelised as a follow-up if needed.

## Verification

Tested against a real AWS S3 bucket in `ca-central-1` with Object Lock Compliance mode (30-day default retention). Before the patch: every multipart upload fails with the InvalidRequest error above. After: uploads complete and objects land under the default retention.

## Related

- #829 — original Object Lock support (PutObject / CopyObject). This PR extends to UploadPart.
- #1324 / #1329 — Content-MD5 on DeleteObjects (related but separate code path).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ vendor/
 .aider*
 AGENTS.md
 CLAUDE.md
+GEMINI.md
 .qwen/
 .claude/
 .crush/
--- a/pkg/storage/s3.go
+++ b/pkg/storage/s3.go
@@ -1,9 +1,12 @@
 package storage
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
+	"encoding/base64"
 	"fmt"
+	"hash/crc32"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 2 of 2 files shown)


david-pivonka:
@Slach when you have a moment, could you take a look at this one? Extends the #829 Object Lock support to the UploadPart path. Happy to adjust the approach (middleware instead of a parallel code path, concurrency, etc.) based on your feedback.

PR #1364, opened 2026-05-05
feat: embed ACVP wrapper and public-scope reproducibility flow

vzakaznikov:
## Summary
- Embed an ACVP wrapper into the shipped `clickhouse-backup` binary with two entry points: argv0 dispatch via `clickhouse-backup-acvp` and CLI dispatch via `clickhouse-backup acvp`.
- Add the `pkg/acvpwrapper` implementation plus tests, and package the `clickhouse-backup-acvp` symlink in Linux package builds so the wrapper can be exercised with `acvptool` against the same binary artifact.
- Add a tracked public-API ACVP config, reproducibility script, and documentation (`pkg/acvpwrapper/acvp_test_fips140v1.26.public.config.json`, `run.sh`, `README.md`) with pinned upstream references and reproduction guidance.
- Keep wrapper capabilities aligned with the supported public Go crypto scope by excluding ML-KEM/ML-DSA from advertised `getConfig` capabilities.

## Test plan
- [x] `docker run --rm -v "$PWD:/work" -w /work golang:1.26-alpine sh -lc 'export PATH=$PATH:/usr/local/go/bin && go test ./...'`
- [x] `bash pkg/acvpwrapper/run.sh` (expected output includes `38 ACVP tests matched expectations`)

Made with [Cursor](https://cursor.com)


Slach:
ACVP test crypto library, and don't test clickhouse-backup itself, let's close this PR

vzakaznikov:
@Slach,  closing is ok. But your provided reason is invalid.

The PR never claimed to test clickhouse-backup's application logic (backup/restore, manifests, storage backends).  It runs ACVP against the cryptographic primitives as compiled into the shipped binary.

In Go, crypto/tls, crypto/aes, crypto/sha*, crypto/ecdsa, crypto/hmac, etc. are statically linked into the
release artifact — there is no external libcrypto at runtime. The bytes ACVP exercises are the exact bytes users
run, and they are the same primitives clickhouse-backup uses for TLS to S3/GCS/Azure/ClickHouse. FIPS 140-3
validation is scoped to the module/binary you ship, which is why the argv0 / clickhouse-backup acvp dispatch
pattern exists.

So "ACVP tests the crypto library, not clickhouse-backup" conflates source-level scope with binary-level scope;
in a statically linked Go binary they are the same code.

Slach:
@vzakaznikov who will run `clickhouse-backup acvp` and why?

vzakaznikov:
> @vzakaznikov who will run `clickhouse-backup acvp` and why?

It is useful to any user of the `clickhouse-backup-fips` binary who wants stronger assurance of
FIPS compatibility than just a reference to Go’s upstream certificate.

The point is not that every user needs to run `clickhouse-backup-fips acvp` during normal operation. The point is
that the exact shipped artifact can be independently exercised with ACVP vectors: same Go version, same
compiler/linker output, same build flags, same packaged binary.

PR #1368, opened 2026-05-10
Prevent nested ssh from consuming while loop stdin

guoard:
Fixed backup iteration over SSH.

The nested `ssh` command was consuming stdin from the `while read` loop, causing only the first backup to be processed. This change uses `ssh -n` to prevent stdin interference and ensures all backup metadata files are processed correctly.


Slach (approved):
nice catch, thanks for contribution

guoard:
Failed test is not related to changes:

```
2026-05-13 18:13:22.546 FTL failed to start test environments error="TestContainers(4).StartAll: start azure: create azure: Error response from daemon: No such image: mcr.microsoft.com/azure-storage/azurite:latest"
```

guoard:
Rebased to master. @Slach PTAL.

PR #1369, opened 2026-05-10
Fix backup retention logic issues for rsync

guoard:
Fix backup retention logic in rsync script by correcting line counting, numeric comparison, and arithmetic handling to ensure proper cleanup of old backups.


guoard:
Rebased to master.

PR #1386, opened 2026-05-17
fix: no retry when remote table metadata not found

### LINKED ISSUE #1379: No retry when table metadata files not exists instead of retrying forever
## Problem

When downloading a backup, `downloadTableMetadata()` retries indefinitely on missing metadata files. If a table was dropped or renamed during upload, its `.json` metadata file may not exist on remote storage, but the retry loop keeps trying with exponential backoff — wasting minutes before ultimately failing the entire download.

### Current behavior

```go
// pkg/backup/download.go — downloadTableMetadata()
retry := retrier.New(retrier.ExponentialBackoff(b.cfg.General.RetriesOnFailure, ...), b)
err := retry.RunCtx(ctx, func(ctx context.Context) error {
    tmReader, err := b.dst.GetFileReader(ctx, remoteMetadataFile)
    if err != nil {
        return errors.Wrapf(err, "can't GetFileReader(%s) error", remoteMetadataFile)
    }
    // ...
})
```

When `GetFileReader` returns "object doesn't exist" / "NoSuchKey" / "StatusCode 404", this error is transient-looking to the retrier, so it retries `RetriesOnFailure` times with exponential backoff. But a 404 is permanent — the object will never appear.

## Proposed Fix

Detect permanent "not found" errors and break out of the retry loop immediately:

```go
retry := retrier.New(retrier.ExponentialBackoff(b.cfg.General.RetriesOnFailure, ...), b)
err := retry.RunCtx(ctx, func(ctx context.Context) error {
    tmReader, err := b.dst.GetFileReader(ctx, remoteMetadataFile)
    if err != nil {
        // "object doesn't exist" is permanent — flag it and stop retrying
        if strings.Contains(err.Error(), "doesn't exist") || 
           strings.Contains(err.Error(), "key not found") || 
           strings.Contains(err.Error(), "NoSuchKey") || 
           strings.Contains(err.Error(), "StatusCode 404") {
            notFoundErr = true
            return nil // break out of retry loop
        }
        return errors.Wrapf(err, "can't GetFileReader(%s) error", remoteMetadataFile)
    }
    // ...
})

// After retry loop:
if notFoundErr {
    log.Warn().Str("remoteMetadataFile", remoteMetadataFile).
        Msg("metadata file not found on remote, skipping table")
    continue
}
```

### When this happens

- Table dropped during backup upload (backup started, table dropped, metadata never uploaded)
- Incremental backup where base backup's metadata was cleaned up
- Partial upload failure where some tables' metadata was never written
- Table renamed between backup create and upload

### Impact

Without this fix, a single missing metadata file causes the entire download to hang for `RetriesOnFailure × exponential_backoff` duration (typically 5-10 minutes) before failing. With this fix, the missing table is skipped in milliseconds and the rest of the backup downloads successfully.

leno23:
## Summary
- detect permanent remote metadata-missing errors before the download metadata retry loop keeps backing off
- warn and skip the affected table when its JSON metadata file is absent on remote storage
- preserve the existing optional `.sql` handling for incremental/embedded metadata and return non-not-found errors directly
- add a focused regression test for the not-found classifier patterns seen from common object stores

## Tests
- `go test ./pkg/backup -run TestIsRemoteMetadataNotFound -count=1`

Closes #1379


Slach:
Changed the missing-`table.json` behavior on `download` from **silent skip** to **fail fast with an error**.

Rationale: a missing table `.json` on remote is a permanent broken-backup condition. Silently skipping it restores fewer tables than the backup metadata claims, with no signal to the operator — a quiet data-loss footgun on restore. A 404 should surface, not be swallowed.

Kept the no-retry fix (the whole point of #1379): the not-found 404 still breaks out of the retry loop immediately instead of burning the ~35s exponential backoff. Detection now covers all backends' phrasings (S3 `NoSuchKey`, GCS `doesn't exist`/`404`, Azure `BlobNotFound`, FTP `No such file or directory`, SFTP `file does not exist`) — without this, FTP/SFTP/Azure were not classified and still retried.

Added per-backend integration tests (`testMetadataNotFound_test.go`, S3/SFTP/FTP/GCS/GCS-emulator/AZBLOB/COS) asserting `download` fails fast with a not-found error and no retry. All 7 green.

Note: the optional `.sql` (incremental/embedded) metadata is still skipped when absent, unchanged.

PR #1387, opened 2026-05-17
fix(resume): validate part completeness before skipping

### LINKED ISSUE #1382: Conservative resume validation: verify part file completeness before skipping
## Problem

The `--resume` flag for both download and upload tracks processed parts in a bbolt state database at the part level. When a download/upload is interrupted mid-part (e.g., 3 of 8 files in a part were transferred), the state records the part as "processed" because the goroutine started it. On resume, the part is skipped entirely — leaving incomplete parts that cause ClickHouse `ATTACH` failures.

### Root cause

The resume state (`resumable.State`) tracks paths via `AppendToState(path, size)` after each part completes. But if the process is killed (OOM, SIGKILL, power loss) between starting a part and finishing all its files, some files exist on disk while others don't. The `.cmrk3` marks files are particularly vulnerable because they're small and written last.

### Symptoms

Download resume:
```
DB::Exception: File data.cmrk3 doesn't exist. (FILE_DOESNT_EXIST)
```

Upload resume:
- Backup appears complete but is missing files for some parts
- Download from this backup fails with missing file errors

### Current behavior (download)

```go
if b.resume {
    isProcessed, pathSize := b.resumableState.IsAlreadyProcessed(partRemotePath)
    if isProcessed {
        atomic.AddUint64(&downloadedSize, uint64(pathSize))
        return nil  // SKIP — but part may be incomplete!
    }
}
```

### Download: validateLocalPartComplete()

Before skipping a "processed" part, validate that ALL expected files actually exist locally with correct sizes:

```go
if b.resume {
    isProcessed, pathSize := b.resumableState.IsAlreadyProcessed(partRemotePath)
    if isProcessed {
        if b.validateLocalPartComplete(dataCtx, partLocalPath, partRemotePath, manifestFiles, partManifestPrefix) {
            atomic.AddUint64(&downloadedSize, uint64(pathSize))
            return nil
        }
        log.Warn().Str("part", partRemotePath).Msg("resume: local part incomplete, re-downloading")
    }
}
```

The validation function:

```go
func (b *Backuper) validateLocalPartComplete(ctx context.Context, partLocalPath, partRemotePath string, 
    manifestFiles []storage.ManifestEntry, partManifestPrefix string) bool {
    // If manifest available, use it as authoritative file list (zero remote calls)
    if len(manifestFiles) > 0 {
        for _, entry := range manifestFiles {
            relativeName := strings.TrimPrefix(entry.Path, partManifestPrefix)
            relativeName = strings.TrimPrefix(relativeName, "/")
            if relativeName == "" { continue }
            localFile := path.Join(partLocalPath, relativeName)
            info, err := os.Stat(localFile)
            if err != nil {
                log.Debug().Str("file", localFile).Err(err).Msg("resume validation: file missing")
                return false
            }
            if info.Size() == 0 && entry.Size > 0 {
                log.Debug().Str("file", localFile).Int64("expected", entry.Size).
                    Msg("resume validation: file is zero-size")
                return false
            }
            if info.Size() != entry.Size {
                log.Debug().Str("file", localFile).Int64("local", info.Size()).
                    Int64("expected", entry.Size).Msg("resume validation: size mismatch")
                return false
            }
        }
        return true
    }

    // No manifest — fall back to remote Walk
    if b.dst == nil { return false }
    var expected []expectedFile
    walkErr := b.dst.Walk(ctx, partRemotePath, true, func(ctx context.Context, f storage.RemoteFile) error {
        expected = append(expected, expectedFile{name: f.Name(), size: f.Size()})
        return nil
    })
    if walkErr != nil { return false }
    // ... verify each expected file exists locally with matching size
}
```

### Upload: validateRemotePartComplete()

Same pattern — before skipping a "processed" part, verify all expected files exist on remote:

```go
if b.resume {
    if isProcessed, processedSize := b.resumableState.IsAlreadyProcessed(remotePathFull); isProcessed {
        if b.validateRemotePartComplete(ctx, remotePathFull, backupPath, partFiles) {
            atomic.AddInt64(&uploadedBytes, processedSize)
            return nil
        }
        log.Warn().Str("part", remotePathFull).Msg("resume: remote part incomplete, re-uploading")
    }
}
```

```go
func (b *Backuper) validateRemotePartComplete(ctx context.Context, remotePath, localBasePath string, 
    expectedFiles []string) bool {
    if b.dst == nil { return false }
    remoteFiles := make(map[string]int64)
    walkErr := b.dst.Walk(ctx, remotePath, true, func(ctx context.Context, f storage.RemoteFile) error {
        remoteFiles[f.Name()] = f.Size()
        return nil
    })
    if walkErr != nil { return false }
    for _, f := range expectedFiles {
        localInfo, err := os.Stat(path.Join(localBasePath, f))
        if err != nil { return false }
        remoteSize, exists := remoteFiles[f]
        if !exists { return false }
        if remoteSize != localInfo.Size() { return false }
    }
    return true
}
```

### Compressed format: StatFile check

For compressed (tar.gz/zstd) uploads, each part is a single archive file. Validate it exists:

```go
if isProcessed, processedSize := b.resumableState.IsAlreadyProcessed(remoteDataFile); isProcessed {
    if _, statErr := b.dst.StatFile(ctx, remoteDataFile); statErr == nil {
        atomic.AddInt64(&uploadedBytes, processedSize)
        return nil
    }
    log.Warn().Str("file", remoteDataFile).Msg("resume: remote archive missing, re-uploading")
}
```

### Key design decisions

- **Conservative**: Any validation failure triggers re-download/re-upload. False negatives (re-downloading a complete part) are safe; false positives (skipping an incomplete part) cause data corruption.
- **Manifest-first**: When a manifest is available, validation uses local `os.Stat` only — zero remote API calls. This makes resume validation near-instantaneous.
- **Nil guards**: Both functions handle nil `b.dst` gracefully (return false = re-transfer).

leno23:
## Summary
- validate resume-marked download parts against the remote file list before skipping them locally
- validate resume-marked uncompressed upload parts against the expected local file list before skipping them remotely
- re-check compressed upload resume entries with `StatFile` before skipping an archive upload
- add regression coverage for missing `.cmrk3`/part files during resume validation

Fixes #1382

## Tests
- `go test ./pkg/backup -run 'TestIs(Local|Remote)PartCompleteRequires' -count=1`
- `go test ./pkg/... -count=1`


Slach:
## Closing: fixes a non-existent bug

Reviewed against `pkg/backup/download.go`, `pkg/backup/upload.go`, and `pkg/resumable/state.go`. The resume mechanism this PR hardens is already correct.

In both transfer modes, `AppendToState(...)` is called **only after** the full transfer returns successfully, so a "processed" entry always implies a completely transferred part/archive:

- **Directory format** (state key = part path): `DownloadPath` / `UploadPath` transfers the whole part directory, then `AppendToState` (`download.go:880-886`, `upload.go:596-603`).
- **Compressed format** (state key = archive file name): `DownloadCompressedStream` / `UploadCompressedStream` (+`StatFile` check on upload), then `AppendToState` (`download.go:797-808`, `upload.go:630-648`).

`AppendToState` commits through a `bolt.DB.Update` transaction (fsync on commit, `state.go:147`). A kill mid-part means the transfer call never returns → no state key → the part is re-transferred on `--resume`. The atomic unit is the whole part (directory) or whole archive (compressed), never a single file.

So `validateLocalPartComplete` / `validateRemotePartComplete` add remote `Walk`/`Stat` work to defend against a state that is never produced. See full analysis in #1382.

Closing along with the issue.

PR #1390, opened 2026-05-19
Improve tables command for per-table backup size breakdown

### LINKED ISSUE #1388: Feature request: per-table backup size breakdown command
## Use case

Before downloading a partial backup with `--table`, there is no way to know how much data you are about to pull. You either download the full backup and check afterward, or guess based on production table sizes (which may differ from the backup).

A new `info` command would show per-table sizes without downloading any data, allowing users to estimate download sizes and plan partial restores.

## Proposed solution

Add a `clickhouse-backup info` command that reads per-table metadata (locally or from remote storage) and displays a size breakdown:

```
clickhouse-backup info [all|local|remote] <backup_name>
clickhouse-backup info -t "default.events*" remote my_backup_2026_05_19
```

### Example output

```
Backup:   my_backup (remote)
Filter:   default.events*

TABLE                  SIZE       PARTS   DISKS
-----                  ----       -----   -----
default.events         1.23TiB    4521    default
default.events_v2      456.78GiB  1203    default
-----                  ----       -----   -----
TOTAL (2 tables)       1.67TiB    5724
```

### Key features

- Support both local and remote backups (reads only metadata, not data)
- `--table` / `-t` flag for glob pattern filtering (reuses existing pattern matching)
- Multiple output formats: text (default), json, yaml, csv, tsv
- For remote backups, only downloads per-table metadata JSON files (fast even for large backups)
- Consistent UX with existing `list` command (same `all|local|remote` selector and `--format` flag)

minguyen9988:
fix https://github.com/Altinity/clickhouse-backup/issues/1388
## Summary

Add a new `info` CLI command that shows per-table size breakdown for local and remote backups.

### Usage
```
clickhouse-backup info [-t, --tables=<db>.<table>] [-f, --format=<text|json|yaml|csv|tsv>] [all|local|remote] <backup_name>
```

### Features
- Shows per-table size, part count, and disk breakdown for any backup
- Supports both local and remote backups (or both with `all`)
- Multiple output formats: text (tabwriter), JSON, YAML, CSV, TSV
- Table pattern filtering with glob wildcards (same syntax as other commands)
- Totals row showing aggregate size and part count

### Files changed
- `cmd/clickhouse-backup/main.go` — Register the `info` command in the CLI (inserted before the `download` command)
- `pkg/backup/info.go` — New file (401 lines) implementing `Info()`, `infoLocal()`, `infoRemote()`, `filterTablesByPattern()`, and `printInfo()`

### Example output
```
Backup: my_backup (local)

TABLE              SIZE       PARTS   DISKS
-----              ----       -----   -----
mydb.events        1.2 GiB    42      default
mydb.logs          256 MiB    12      default
-----              ----       -----   -----
TOTAL (2 tables)   1.5 GiB    54
```


minguyen9988:
https://github.com/Altinity/clickhouse-backup/issues/1388

Slach (changes requested):
we already have "tables" command which cover 80% of your implemented functions
look
`clickhouse-backup tables --help`

please modify exists code instead of implement duplicated functions

minguyen9988:
the --tables does not work on local backup it is only listing table in the clickhouse server and there is an option to list table in remote backup, there can be many local backup covering different tables and the tables command doesnt follow the convention in the `list` command, 

the table command is actually quite confusing, I suggest we can remove the --remote-backup option and use the info command instead,

The table is suppose to list table in a backup or in the clickhouse server in this case. I think it should be used for listing table in the clickhouse-server that is eligible for backup given the config (remove the skip table).

```
$ clickhouse-backup tables --help
NAME:
   clickhouse-backup tables - List of tables, exclude skip_tables

USAGE:
   clickhouse-backup tables [--tables=<db>.<table>] [--remote-backup=<backup-name>] [--all]

OPTIONS:
   --config value, -c value                   Config 'FILE' name. (default: "/etc/clickhouse-backup/config.yml") [$CLICKHOUSE_BACKUP_CONFIG]
   --environment-override value, --env value  override any environment variable via CLI parameter
   --all, -a                                  Print table even when match with skip_tables pattern
   --table value, --tables value, -t value    List tables only match with table name patterns, separated by comma, allow ? and * as wildcard
   --remote-backup value                      List tables from remote backup

```

the info command is cleaner and on point I think 

```
$ clickhouse-backup info --help
NAME:
   clickhouse-backup-bin info - Show per-table size breakdown for a backup

USAGE:
   clickhouse-backup info [-t, --tables=<db>.<table>] [-f, --format=<text|json|yaml|csv|tsv>] [all|local|remote] <backup_name>

OPTIONS:
   --config value, -c value                   Config 'FILE' name. (default: "/etc/clickhouse-backup/config.yml") [$CLICKHOUSE_BACKUP_CONFIG]
   --environment-override value, --env value  override any environment variable via CLI parameter
   --table value, --tables value, -t value    Show info only for matched table name patterns, separated by comma, allow ? and * as wildcard
   --format value, -f value                   Output format (text|json|yaml|csv|tsv)
   
```

Slach:
Thanks for the explanation, but the duplication is still real and I'd like to keep one command instead of two. Concretely:

1. `GetTablesRemote` (pkg/backup/list.go:585) already lists tables from a remote backup; we should enrich it with size/parts (the data is in `metadata.TableMetadata`) rather than re-implementing the BackupDestination → BackupList → per-table metadata download path in a new file.
2. The "list tables from a local backup" use case (no live ClickHouse needed) is the only thing genuinely missing from `tables` — add it there as a symmetric `--local-backup=<name>` flag using the same `metadata.TableMetadata.Load(...)` you already use in `info.go`.
3. The `--format=text|json|yaml|csv|tsv` switch is a generally useful flag — please attach it to `tables` (and ideally to `list` later in a follow-up) instead of inventing a new command for it.
4. The new command is wired only into the CLI. `pkg/server/server.go` already exposes `/backup/tables` and `/backup/tables/all` (server.go:239-240, handler at 778-802). Whatever lives in the CLI must also be reachable via the REST API — add a `local_backup` query parameter to `httpTablesHandler` so the HTTP surface stays at feature parity with the CLI.
5. Coverage shows 0/298 lines covered in `pkg/backup/info.go` and 0/10 in the `cmd/...` diff. The only test change is the CLI snapshot (`cli.py.cli.snapshot`), which just reflects that the command shows up in `--help` — that is not a functional test. Please add an integration test at `test/integration/tables_test.go` that exercises the new flags for both local and remote backups.

I'm closing this in favor of folding the useful parts (local-backup reader, format outputs, size/parts for remote) into the existing `tables` command with a proper REST endpoint and an integration test. Apologies for the back-and-forth, and thanks for the input on what `tables` was missing — that part was a fair point.

minguyen9988:
yes, I think if you want to merge and add more features to the tables command that is good as well, please take over, thanks

Slach (approved):

PR #1391, opened 2026-05-19
Added QA_STP_ClickHouse_Backup_FIPS.md

besteffects:
Added test plan for clickhouse-backup compatibility with fips140


Slach (approved):
ok. will wait when master will stable then will merge it

PR #1397, opened 2026-05-26
fix: prefer underlying disk name over cache wrapper in getDisksFromSystemDisks

### LINKED ISSUE #1396: Backup fails with 'part not found after FREEZE' when cache disk wraps an S3 object disk
# Bug: Backup fails with "part not found after FREEZE" when cache disk wraps an S3 object disk

## Summary

`clickhouse-backup` fails to back up tables stored on S3 object disks when a ClickHouse cache disk (e.g., `s3_cache`) wraps the underlying object disk (e.g., `s3_disk`). The error is:

```
part "<part_name>" not found in system.parts (database=<db>, table=<table>, disk_name=s3_cache) after FREEZE
```

## Environment

- **ClickHouse version:** 25.11.8.25
- **clickhouse-backup version:** v2.7.0
- **Storage configuration:** Hybrid tiered storage with EBS (local) + S3 object disk + S3 cache disk
- **Architecture:** linux/arm64

## Storage Configuration

```xml
<storage_configuration>
  <disks>
    <default>
      <keep_free_space_bytes>1073741824</keep_free_space_bytes>
    </default>
    <s3_disk>
      <type>s3</type>
      <endpoint>https://bucket.s3.us-east-1.amazonaws.com/prefix/</endpoint>
      <use_environment_credentials>true</use_environment_credentials>
      <metadata_path>/var/lib/clickhouse/disks/s3_disk/</metadata_path>
    </s3_disk>
    <s3_cache>
      <type>cache</type>
      <disk>s3_disk</disk>
      <path>/var/lib/clickhouse/filesystem_caches/s3_cache/</path>
      <max_size>10737418240</max_size>
    </s3_cache>
  </disks>
  <policies>
    <s3_tiered>
      <volumes>
        <main><disk>default</disk></main>
        <cold><disk>s3_cache</disk></cold>
      </volumes>
    </s3_tiered>
  </policies>
</storage_configuration>
```

## Root Cause

In `getDisksFromSystemDisks()` ([pkg/clickhouse/clickhouse.go](https://github.com/Altinity/clickhouse-backup/blob/v2.7.0/pkg/clickhouse/clickhouse.go#L307)), the SQL query uses `GROUP BY d.path` with `any(d.name)` to deduplicate disks by path. However, when a cache disk wraps an underlying object disk, **both disks share the same `d.path`** (metadata path):

```sql
SELECT name, path, type, cache_path FROM system.disks;
```

| name      | path                               | type          | cache_path                                     |
|-----------|------------------------------------|--------------|-------------------------------------------------|
| default   | /var/lib/clickhouse/               | Local         |                                                 |
| s3_cache  | /var/lib/clickhouse/disks/s3_disk/ | ObjectStorage | /var/lib/clickhouse/filesystem_caches/s3_cache/ |
| s3_disk   | /var/lib/clickhouse/disks/s3_disk/ | ObjectStorage |                                                 |

Because both `s3_cache` and `s3_disk` share the same `d.path`, `GROUP BY d.path` collapses them into a single row. The `any(d.name)` aggregate function then **arbitrarily** picks either `s3_cache` or `s3_disk`.

When `s3_cache` is selected, `clickhouse-backup` later queries `system.parts` using `disk_name = 's3_cache'`. However, **`system.parts` always reports the underlying disk name** (`s3_disk`), not the cache wrapper. This mismatch causes `fetchHashOfAllFiles` to find zero parts and fail with:

```
part "all_1_1_0" not found in system.parts (database=mydb, table=my_table, disk_name=s3_cache) after FREEZE
```

### Why `d.type` cannot distinguish cache disks

In modern ClickHouse versions (25.x+), both the cache disk and its underlying disk report `type = 'ObjectStorage'` in `system.disks`. The older `type = 'Cache'` classification is no longer used. This means filtering on `d.type` is insufficient to distinguish cache disks from their underlying disks.

### Evidence from `system.disks`

```sql
SELECT name, path, type, object_storage_type, cache_path FROM system.disks;
```

| name      | type          | object_storage_type | cache_path                                     |
|-----------|--------------|---------------------|-------------------------------------------------|
| default   | Local         | None                |                                                 |
| s3_cache  | ObjectStorage | S3                  | /var/lib/clickhouse/filesystem_caches/s3_cache/ |
| s3_disk   | ObjectStorage | S3                  | *(empty)*                                       |

The **`cache_path` column** is the only reliable differentiator: cache disks have a non-empty `cache_path`, while underlying disks have an empty `cache_path`.

## Steps to Reproduce

1. Configure ClickHouse with a tiered storage policy using an S3 object disk + cache disk (as shown above)
2. Create a table using the `s3_tiered` storage policy
3. Insert data so that some parts are on the `default` (EBS) disk and some on `s3_disk` (cold tier)
4. Run `clickhouse-backup create` or `clickhouse-backup create_remote`
5. Observe the error:
   ```
   part "..." not found in system.parts (database=..., table=..., disk_name=s3_cache) after FREEZE
   ```

Note: The error is **non-deterministic** — it depends on which disk name `any(d.name)` happens to pick. In some cases it may pick `s3_disk` (correct) and the backup succeeds; in others it picks `s3_cache` (incorrect) and the backup fails.

## Expected Behavior

`clickhouse-backup` should always resolve the underlying (non-cache) disk name when cache disks are present, since `system.parts` reports the underlying disk name.

## Actual Behavior

`clickhouse-backup` may resolve the cache disk name (`s3_cache`) instead of the underlying disk name (`s3_disk`), causing a mismatch with `system.parts` and a backup failure.

## Workaround

There is no workaround without patching the binary. Setting `disk_mapping` does not help because it adds the cache disk as a separate entry rather than remapping the name.

vizsmurali:
# fix: prefer underlying disk name over cache wrapper in getDisksFromSystemDisks

## Problem

When a ClickHouse cache disk wraps an S3 object disk, both disks share the same
`path` (metadata path) in `system.disks`. The `GROUP BY d.path` clause in
`getDisksFromSystemDisks()` collapses them into one row, and `any(d.name)` may
arbitrarily pick the cache disk name (e.g., `s3_cache`) instead of the
underlying disk name (e.g., `s3_disk`).

This causes `fetchHashOfAllFiles` to query `system.parts` with
`disk_name = 's3_cache'`, but `system.parts` always reports the underlying disk
name (`s3_disk`), resulting in:

```
part "..." not found in system.parts (database=..., table=..., disk_name=s3_cache) after FREEZE
```

The bug is **non-deterministic** — `any()` may pick either name depending on
the query execution order.

### Why `d.type` is insufficient

In ClickHouse 25.x+, both cache and underlying disks report
`type = 'ObjectStorage'` in `system.disks`. The older `type = 'Cache'`
classification is no longer used, so type-based filtering cannot distinguish
them.

## Fix

Use the `cache_path` column from `system.disks` to reliably distinguish cache
disks from underlying disks:

- Cache disks have a **non-empty** `cache_path`
  (e.g., `/var/lib/clickhouse/filesystem_caches/s3_cache/`)
- Underlying disks have an **empty** `cache_path`

Replace `any(d.name)` with `argMin(d.name, d.cache_path != '')`:

- For non-cache disks: `cache_path != ''` evaluates to `0` (false)
- For cache disks: `cache_path != ''` evaluates to `1` (true)
- `argMin` picks the name with the **minimum** sort value, i.e., the
  non-cache disk name (`0 < 1`)

The fix also detects whether the `cache_path` column exists in `system.disks`
(for backward compatibility with older ClickHouse versions) before using it.

## Changes

- `pkg/clickhouse/clickhouse.go`:
  - Added `CachePathPresent` field to `DiskFields` struct
  - Added detection of `cache_path` column in `system.columns`
  - Changed disk name SQL from `any(d.name)` to
    `argMin(d.name, d.cache_path != '')` when `cache_path` column exists
  - Parameterized the disk name SQL in the final query

### `system.disks` showing the shared path problem

```
┌─name─────┬─path───────────────────────────────┬─type──────────┬─cache_path──────────────────────────────────┐
│ default  │ /var/lib/clickhouse/               │ Local         │                                             │
│ s3_cache │ /var/lib/clickhouse/disks/s3_disk/ │ ObjectStorage │ /var/lib/clickhouse/filesystem_caches/s3_cache/ │
│ s3_disk  │ /var/lib/clickhouse/disks/s3_disk/ │ ObjectStorage │                                             │
└──────────┴────────────────────────────────────┴───────────────┴─────────────────────────────────────────────┘
```

### Before fix (v2.7.0)

```
ERR part "all_1_42_7" not found in system.parts
    (database=mydb, table=my_table, disk_name=s3_cache) after FREEZE
```

### After fix

Successfully backed up a database with multiple tables (TiB-scale S3 object disk
+ GiB-scale EBS local data) with zero errors on a ClickHouse 25.11 cluster.

## Backward Compatibility

- The `cache_path` column detection is gated behind a check in
  `system.columns`, so older ClickHouse versions without this column will
  continue to use `any(d.name)` as before.
- When no cache disks are configured, `argMin(d.name, d.cache_path != '')`
  degrades to picking an arbitrary name (all have `cache_path = ''`), which is
  equivalent to the current `any(d.name)` behavior.

## Testing

- Manually tested on ClickHouse 25.11 with hybrid tiered storage
  (local disk + S3 object disk + S3 cache disk)
- Full database backup of multiple tables with TiB-scale S3 object disk data,
  completed with zero errors
- Full database restore: verified disk routing (hot data → local, cold data → S3)

Fixes #1396

### DIFF
--- a/pkg/backup/create.go
+++ b/pkg/backup/create.go
@@ -1081,11 +1081,20 @@ func (b *Backuper) AddTableToLocalBackup(ctx context.Context, backupName string,
 	return disksToPartsMap, realSize, objectDiskSize, checksums, hashOfAllFiles, nil
 }
 
-// fetchHashOfAllFiles returns name → hash_of_all_files for the given parts on
-// disk `diskName`. The value is whatever ClickHouse prints in
-// system.parts.hash_of_all_files (already lowercased); the storage and
-// compare paths are version-agnostic because both ends use the server's
-// formatting.
+// fetchHashOfAllFiles returns name → hash_of_all_files for the given parts of
+// `database`.`table` (frozen from disk `diskName`). The value is whatever
+// ClickHouse prints in system.parts.hash_of_all_files (already lowercased); the
+// storage and compare paths are version-agnostic because both ends use the
+// server's formatting.
+//
+// We deliberately do NOT filter by disk_name: when a cache disk wraps an object
+// disk, both share the same metadata path, and system.parts reports the part
+// under the cache disk name (the disk named in the storage policy), not the
+// underlying disk that clickhouse-backup walks the shadow tree from. Filtering
+// by diskName would miss those parts (see issue #1396). Part names are unique
+// within a table's active set, and inactive copies left by a move carry
+// identical content (hence identical hash_of_all_files), so matching by
+// database+table+name is unambiguous.
 //
 // Called right after FREEZE so the active-set delta is essentially zero —
 // inactive parts also remain visible in system.parts for ~480s, which makes
@@ -1096,8 +1105,8 @@ func (b *Backuper) fetchHashOfAllFiles(ctx context.Context, database, table, dis
 		Name string `ch:"name"`
 		Hash string `ch:"hash_of_all_files"`
 	}
-	q := "SELECT name, lower(hash_of_all_files) AS hash_of_all_files FROM system.parts WHERE database=? AND `table`=? AND disk_name=? AND has(?, name)"
-	if err := b.ch.SelectContext(ctx, &rows, q, database, table, diskName, partNames); err != nil {
+	q := "SELECT name, lower(hash_of_all_files) AS hash_of_all_files FROM system.parts WHERE database=? AND `table`=? AND has(?, name)"
+	if err := b.ch.SelectContext(ctx, &rows, q, database, table, partNames); err != nil {
 		return nil, errors.Wrap(err, "SELECT hash_of_all_files FROM system.parts")
 	}
 	hashByName := make(map[string]string, len(rows))
--- a/pkg/backup/create.go
+++ b/pkg/backup/create.go
@@ -1099,7 +1099,9 @@ func (b *Backuper) AddTableToLocalBackup(ctx context.Context, backupName string,
 // Called right after FREEZE so the active-set delta is essentially zero —
 // inactive parts also remain visible in system.parts for ~480s, which makes
 // the race window practically impossible to hit. If a frozen part is missing
-// from the result we surface a hard error rather than silently dropping it.
+// from the result we surface a hard error rather than silently dropping it,
+// unless the table itself was dropped/detached concurrently (the documented
+// IgnoreNotExistsErrorDuringFreeze race), in which case we skip the part.
 func (b *Backuper) fetchHashOfAllFiles(ctx context.Context, database, table, diskName string, partNames []string) (map[string]string, error) {
 	var rows []struct {
 		Name string `ch:"name"`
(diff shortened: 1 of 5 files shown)


Slach (changes requested):
could you add e2e test in test/integration/ ?

vizsmurali:
Thanks for the review! I've added an e2e integration test in `test/integration/cacheDiskBackup_test.go` (`TestCacheDiskBackup`).

The test:
1. Configures ClickHouse with `s3_disk` + `s3_cache` (cache wrapper) + `s3_tiered` storage policy
2. Creates a table with TTL to move old data to the S3 cold volume
3. Runs backup — this is where the bug manifested (`part not found after FREEZE`)
4. Uploads, downloads, and restores the backup
5. Verifies data integrity (row count) after restore

Requires ClickHouse >= 22.3 for cache disk support. Compiles and passes `go vet`.

@Slach please review
Thanks again

Slach (approved):

Slach:
@vizsmurali thanks for contribution

PR #1407, opened 2026-06-03
fix: don't abort backup when resumable state DB read/write fails (#1172)

### LINKED ISSUE #1172: replace Fatal to Warning when failed to write to resumable state
need survive during following errors
`2025-06-14 02:43:06.649 FTL pkg/resumable/state.go:152 > resumable state: can't write key shard1-20250614000119/shadow/default/hits_raw/default_20250305_145983_165220_26.tar to /var/lib/clickhouse/backup/shard1-20250614000119/upload.state2 error: write /var/lib/clickhouse/backup/shard1-20250614000119/upload.state2: no space left on device`

mvanhorn:
## Summary

Replaces the three `log.Fatal().Stack()` calls in `pkg/resumable/state.go` with `log.Warn()`, so a failure to read or write the resumable state DB no longer calls `os.Exit(1)` and aborts an in-progress operation. Adds nil-bucket guards inside the bolt closures so removing the fatals does not expose a nil-pointer panic, and adds unit tests for the read/write error paths.

## Why this matters

Per issue #1172, when the resumable state DB cannot be written, the process aborts. The motivating case is a disk-full condition:

```
FTL pkg/resumable/state.go:152 > resumable state: can't write key shard1-.../upload.state2 error: write .../upload.state2: no space left on device
```

The resumable state is an optimization (skip already-processed parts on retry). Losing the ability to record it should degrade gracefully, not kill the whole backup/upload/download/restore. Every other error path in this file already uses `log.Warn()` (open, loadParams, cleanup, saveParams, Close), so the three `Fatal` calls were inconsistent with the file's own conventions.

After this change:
- `AppendToState`: a write error logs a warning and returns; the run continues.
- `IsAlreadyProcessed`: a read error logs a warning and returns `(false, 0)`, so the affected part is simply re-processed.
- `getBucket`: a missing bucket logs a warning and returns the (nil) bucket; the callers' closures early-return on nil to avoid a panic.

## Testing

- `gofmt -l pkg/resumable/` — clean
- `go vet ./pkg/resumable/...` — clean
- `go build ./...` — passes
- `go test ./pkg/resumable/...` — passes, including new tests that close the underlying DB to force the read/write error paths and assert the process survives (`TestAppendToStateSurvivesWriteError`, `TestIsAlreadyProcessedSurvivesReadError`) plus a happy-path round-trip (`TestAppendThenIsAlreadyProcessed`).

Fixes #1172

AI was used for assistance.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,5 @@
 # v2.7.1
-- don't abort backup/upload/download/restore when the resumable state DB can't be written or read (e.g. `no space left on device`); log a warning and continue, fix [#1172](https://github.com/Altinity/clickhouse-backup/issues/1172)
+- don't kill `clickhouse-backup server` with `Fatal`/`os.Exit` when the resumable state DB can't be written or read (e.g. `no space left on device`); the error now propagates so the server stays alive and returns it to the API client, while the CLI exits with a non-zero code, fix [#1172](https://github.com/Altinity/clickhouse-backup/issues/1172)
 
 # v2.7.0
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,6 @@
+# v2.7.1
+- don't abort backup/upload/download/restore when the resumable state DB can't be written or read (e.g. `no space left on device`); log a warning and continue, fix [#1172](https://github.com/Altinity/clickhouse-backup/issues/1172)
+
 # v2.7.0
 
 NEW FEATURES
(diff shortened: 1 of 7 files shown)


Slach:
Thanks a lot for tackling this, @mvanhorn! 🙏

I reworked the fix to match what "surviving" was meant to be in #1172: instead of swapping the three `log.Fatal()` calls for `log.Warn()` and silently continuing, `AppendToState` / `IsAlreadyProcessed` / `IsAlreadyProcessedBool` now **return the error** up the call stack. That way the long-running `clickhouse-backup server` no longer dies via `os.Exit(1)` — it stays alive and returns the error to the API client, while the CLI exits with a non-zero code. The error is now propagated through every caller in create/upload/download/restore (including the errgroup goroutines), and the unit tests were updated to assert the error is returned rather than swallowed.

Pushed onto your branch. Appreciate the clean reproduction and the nil-bucket guards — kept those.

mvanhorn:
Thanks for the merge @Slach! Backups surviving resumable-state DB hiccups feels like the right default.

mvanhorn:
Thanks @Slach for merging this. Backups surviving a flaky resumable-state DB should save some 3am pages.

PR #1430, opened 2026-06-19
fix(restore): eliminate O(N^2) linear scan in getTableListByPatternLocal and getTableListByPatternRemote

ruslanen:
addTableToListIfNotExistsOrEnrichQueryAndParts iterated the full result slice for every file processed during filepath.Walk. With 550K tables this produced N*(N-1)/2 ~ 151 billion string comparisons (~2 hours).

Fix: maintain a tableIndex map[TableTitle]int alongside the result slice. Lookup and enrich are now O(1); total restore init phase drops from O(N^2) to O(N). The inline addToResult closure replaces the two local call sites; the original function is kept for the remote-restore path which operates on far fewer tables in practice.


Slach:
Follow-up commit: pushed the O(1) \`tableIndex\` into \`addTableToListIfNotExistsOrEnrichQueryAndParts\` instead of an inline closure, so `getTableListByPatternLocal` and `getTableListByPatternRemote` share one implementation and both run O(N) (the remote path was still O(N²)). Added a unit test for the .sql/.json merge, no-duplicate, index correctness, and the non-empty-field-not-overwritten invariant.

PR #1431, opened 2026-06-19
Fix/batch inprogress mutations

ruslanen:
## Batch system.mutations lookup to avoid O(N²) on installations with many tables

  ### Problem
  During `create`, in-progress mutations are fetched **once per table** via
  `GetInProgressMutations(ctx, database, table)`:

  ```sql
  SELECT mutation_id, command FROM system.mutations WHERE is_done=0 AND database=? AND table=?
  ```

  `system.mutations` is a virtual table: **every** query against it enumerates all tables on the
  server, so the per-table `WHERE database/table` filter does not bound the work that ClickHouse
  actually does. With N tables in the backup we issue N such queries, and each one costs
  ~O(total tables) → overall **O(N²)**.

  On installations with many tables this single query family dominates `create` wall-clock. Observed
  on a real cluster: ~240 ms per call across tens of thousands of tables, collapsing `create`
  throughput from ~250 tables/s to ~26 tables/s.

  ### Fix
  Fetch the whole in-progress mutation set **once per backup** with a single scan:

  ```sql
  SELECT database, table, mutation_id, command FROM system.mutations WHERE is_done=0
  ```

  `GetInProgressMutationsBatch` returns a `map["database.table"][]MutationMetadata`; the per-table
  code path now does an in-memory map lookup instead of a query. Query count drops from **N → 1**.

  ### Behavior unchanged
  - Same `WHERE is_done=0` filter as before.
  - Same per-table `Mutations` written into `TableMetadata`.
  - The batch query runs only when the per-table query would have run before (`BackupMutations`
    enabled and not schema/rbac/configs/named-collections-only).

  ### Tests
  Added `TestGroupMutationsByTable` (+ empty-input case) covering the pure `groupMutationsByTable`
  helper: rows from one server-wide scan are bucketed to the correct `database.table` with no
  cross-table leakage and in stable order.

  ### Impact (measured)
  55k-table / 93.6 GiB local backup: `create` ~35 min → ~230 s.

  ### Files
  - `pkg/clickhouse/clickhouse.go` — `GetInProgressMutationsBatch`, `inProgressMutationRow`, `groupMutationsByTable`
  - `pkg/backup/create.go` — single batch call, per-table map lookup
  - `pkg/clickhouse/clickhouse_test.go` — unit tests

### DIFF
--- a/pkg/clickhouse/clickhouse.go
+++ b/pkg/clickhouse/clickhouse.go
@@ -1364,11 +1364,12 @@ func (ch *ClickHouse) GetInProgressMutations(ctx context.Context, database strin
 }
 
 // GetInProgressMutationsBatch returns all in-progress mutations across the whole server in a
-// SINGLE query, keyed by "database.table". system.mutations is an expensive virtual table — every
-// query against it enumerates all tables on the server — so calling GetInProgressMutations once per
-// table is O(N^2) and dominates `create` wall-clock on installations with many tables (observed:
-// ~240ms/call * tens of thousands of tables). Fetching the whole in-progress set once per backup
-// turns that into a single O(N) scan; per-table lookup is then an in-memory map access.
+// SINGLE query, keyed by metadata.TableTitle{Database, Table}. system.mutations is an expensive
+// virtual table — every query against it enumerates all tables on the server — so calling
+// GetInProgressMutations once per table is O(N^2) and dominates `create` wall-clock on
+// installations with many tables (observed: ~240ms/call * tens of thousands of tables). Fetching
+// the whole in-progress set once per backup turns that into a single O(N) scan; per-table lookup is
+// then an in-memory map access.
 type inProgressMutationRow struct {
 	Database   string `ch:"database"`
 	Table      string `ch:"table"`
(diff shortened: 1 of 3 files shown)


pkg/clickhouse/clickhouse.go
```diff
@@ -1363,6 +1363,39 @@ func (ch *ClickHouse) GetInProgressMutations(ctx context.Context, database strin
...
+	MutationId string `ch:"mutation_id"`
+	Command    string `ch:"command"`
+}
+
+// groupMutationsByTable buckets flat mutation rows into per-table lists keyed by
+// "database.table". Pure (no I/O) so it is unit-testable without a ClickHouse server.
+func groupMutationsByTable(rows []inProgressMutationRow) map[string][]metadata.MutationMetadata {
+	result := make(map[string][]metadata.MutationMetadata, len(rows))
+	for _, r := range rows {
+		key := r.Database + "." + r.Table
```
Slach: look to metadata.TableTitle 

r.Database + "." + r.Table

this is bad key

you not cover lot of corner cases

Slach (changes requested):

pkg/clickhouse/clickhouse.go
```diff
@@ -1363,6 +1363,39 @@ func (ch *ClickHouse) GetInProgressMutations(ctx context.Context, database strin
...
+		result[key] = append(result[key], metadata.MutationMetadata{MutationId: r.MutationId, Command: r.Command})
+	}
+	return result
+}
+
+func (ch *ClickHouse) GetInProgressMutationsBatch(ctx context.Context) (map[string][]metadata.MutationMetadata, error) {
+	var rows []inProgressMutationRow
+	query := "SELECT database, table, mutation_id, command FROM system.mutations WHERE is_done=0"
```
Slach, line 1395: maybe better add filter here for tables and databases which we already try to backup

Slach (changes requested):
thanks for contribution

but, actually if you have more than 1000 mutations, this is means something wrong with your achitecture

scan in-memory system.mutations for each table, is not so dangerous

Slach:
ok. i figure out about O(N^2) for system.mutation, you are right let me fix TableTitle approach

Slach (approved):

PR #1441, opened 2026-06-26
fix: mask sensitive values in env override logging

### LINKED ISSUE #1429: --env overrides show values of potentially sensitive variables
hide sensitive values (PASSWORD/SECRET_KEY/ACCESS_KEY/...) in override env logging

mvanhorn:
## Summary

When environment variables are overridden via `--env`, the override is logged at info level as `override NAME=VALUE` with the value in cleartext. For secret-bearing variables (passwords, access keys, storage credentials, encryption keys) that writes the raw secret into the log, where it can persist in log files, aggregators, or CI output. This masks the value in that log line when the variable name looks sensitive, printing `[MASKED]` instead, while leaving non-sensitive overrides readable.

## Why this matters

Issue #1429 asks for the env-override logging to stop leaking secrets. The override log is otherwise useful for debugging which variables were applied, so the fix keeps the line but redacts only the value, and only when the variable name matches a sensitive token. The token list covers the documented credential and encryption-key variables across backends (passwords, secrets, access/secret keys, `AZBLOB_ACCOUNT_KEY` / SSE keys, `S3_SSE_CUSTOMER_KEY`, `GCS_ENCRYPTION_KEY`, tokens, SAS, connection strings), so common Azure/GCS/S3 SSE configurations no longer expose credentials in logs.

## Changes

`pkg/config/config.go` gains a `maskSensitiveEnvValue(name, value)` helper backed by a `sensitiveEnvNameTokens` list and uses it only at the `override NAME=VALUE` info-log call site. The override behavior itself is unchanged; only the logged representation of the value changes for sensitive names.

## Testing

Added tests in `pkg/config/config_test.go` asserting that sensitive overrides (including the Azure/GCS/SSE key variables) log `[MASKED]` and never leak the raw value, that a non-sensitive variable still logs its value, and that the env override is applied regardless. The log-capture test pins `LOG_LEVEL=info` so it is independent of the surrounding process's log level. `go test ./pkg/config` passes (including under `LOG_LEVEL=error`); `go vet` and `gofmt` are clean.

Closes #1429


Slach (approved):

Slach (approved):

Slach:
@mvanhorn thanks for contribution!

Issue #1376, opened 2026-05-17
Optimize buffer sizes and HTTP transport for high-bandwidth S3/GCS transfers


minguyen9988:
## Problem

The default buffer sizes and HTTP transport settings in clickhouse-backup are tuned for low-bandwidth environments. On modern 10Gbps+ networks with high-latency object storage (S3, GCS, MinIO), these defaults severely limit throughput.

### Default bottlenecks

1. **PipeBufferSize (ring buffer between compression and upload):** 128KB — causes frequent context switches between producer/consumer goroutines
2. **S3 BufferSize (multipart upload buffer):** default from aws-sdk — results in many small PutObject calls
3. **S3 multipart chunk size:** 5MB default — creates too many parts for large files, each with its own HTTP round-trip
4. **HTTP transport connection pool:** Go's default `MaxIdleConnsPerHost=2` — serializes parallel uploads/downloads that target the same S3 endpoint
5. **io.Copy buffer:** Go's default 32KB — excessive syscalls per file transfer
6. **GCS PutFile buffer:** small default — frequent flushes to the GCS API

### Impact

On a 10Gbps link to S3-compatible storage (MinIO), we measured ~200-400 MB/s throughput with defaults. After tuning, throughput reached 800-1200 MB/s — a 3-4x improvement with zero code logic changes.

### 1. Increase PipeBufferSize from 128KB to 8MB

```go
// pkg/storage/general.go
const (
    // PipeBufferSize - size of ring buffer between stream handlers
    PipeBufferSize = 8 * 1024 * 1024  // was 128 * 1024
)
```

Larger ring buffers reduce context-switch overhead between the compression goroutine and the upload goroutine. The 8MB size matches the typical L3 cache line and allows compression to run ahead of uploads without blocking.

### 2. Tune S3 HTTP transport for high concurrency

```go
// pkg/storage/s3.go — in NewS3() or equivalent init
httpTransport := &http.Transport{
    MaxIdleConns:          512,           // was default 100
    MaxIdleConnsPerHost:   128,           // was default 2 (!)
    MaxConnsPerHost:       0,             // unlimited
    IdleConnTimeout:       120 * time.Second,
    TLSHandshakeTimeout:  10 * time.Second,
    ExpectContinueTimeout: 1 * time.Second,
    WriteBufferSize:       1 * 1024 * 1024, // 1MB (was 4KB)
    ReadBufferSize:        1 * 1024 * 1024, // 1MB (was 4KB)
[...]

PR #1491, opened 2026-08-01
fix(s3): add HTTP/2 keep-alive and write timeouts to avoid wedged requests (#1490)

### LINKED ISSUE #1490: Deadlock in object-disk server-side copy during incremental backup
**Version:** v2.8.0 (`4598867`), ClickHouse 25.8, S3 remote + S3 object disk
(`object_disk_server_side_copy: true`), ~3200 object-disk parts.

### Problem
An incremental `create_remote` of a DB with a large S3 object disk **deadlocks**
during the object-disk upload phase: goroutines stay alive but there is no
progress, no S3 traffic, and `done, duration` is never printed. In `watch` mode
the whole daemon wedges. It's **intermittent** (one backup succeeds, the next on
the same data hangs).

### Root cause
Two nested errgroups both use `SetLimit`, and the inner `Go()` blocks forever on
the limit semaphore while parents wait:

- `uploadObjectDiskParts` — `pkg/backup/create.go:1340`
  `SetLimit(ObjectDiskServerSideCopyConcurrency)` (default **32**), then
  `.Go(func(){ ... s3.CopyObject() ... })`
- `S3.CopyObject` — `pkg/storage/s3.go:996`
  `SetLimit(s.Config.Concurrency)` (default **4**), loop of `.Go()` (`s3.go:1005`)
  then `.Wait()` (`s3.go:1035`)

Up to 32 `CopyObject` run at once, each opening its own limited inner errgroup.
Goroutine dump while hung (503 goroutines):

 52 [chan send]          errgroup.(*Group).Go        errgroup.go:74
                         S3.CopyObject               pkg/storage/s3.go:1005
                         uploadObjectDiskParts...     pkg/backup/create.go:1381
  2 [sync.WaitGroup.Wait] errgroup.(*Group).Wait      errgroup.go:56
                         S3.CopyObject               pkg/storage/s3.go:1035

`errgroup`'s `SetLimit` acquire (`g.sem <- token`) ignores context cancellation,
so once the semaphore saturates the blocked `Go()` calls never return — permanent
deadlock.

### Reproduce
DB with tables on an S3 object disk (a few thousand parts) → base backup →
`create_remote --diff-from-remote=<base> <new>`. Hangs within minutes;
intermittent.

Alexays:
### Problem

Fixes #1490.

The S3 client sets **no per-request timeout**. When a connection stops making progress — a throttling / back-pressuring endpoint (e.g. Scaleway Object Storage) that stops reading the request or sending a response — an in-flight request wedges forever with no way to recover.

During object-disk **server-side copy** this is very visible: dozens of concurrent `UploadPartCopy` requests are multiplexed over a few HTTP/2 connections, and once one stalls, the goroutines pile up in `net/http.(*http2clientStream).writeRequest` (serialized per connection). A goroutine dump from a wedged `create_remote` showed ~218 requests stuck there, 0 network, and the backup never progressing (in `watch` mode the whole daemon wedges). It reproduces intermittently under load and never self-recovers.

`http_idle_conn_timeout` doesn't help here — it only covers *idle* connections, not an active request whose write/response has stalled.

### Fix

Always build a custom transport and configure the HTTP/2 transport with health checks and a write timeout:

- `ReadIdleTimeout = 30s` + `PingTimeout = 15s` — PING the peer when the connection goes idle; if there's no response, the connection is closed.
- `WriteByteTimeout = 60s` — close the connection if a single write stalls for too long (this is exactly the `writeRequest` stall above).

When a connection stalls it is now dropped, the in-flight request fails, and it is retried on a fresh connection instead of blocking indefinitely. Validated against Scaleway Object Storage: object-disk backups that previously wedged indefinitely now complete.

The timeouts are conservative constants here for a minimal change; happy to expose them as `s3` config options if you'd prefer.


Slach (changes requested):
Thanks — the diagnosis in #1490 holds up (218 goroutines parked in `http2clientStream.writeRequest`, never recovering), and dropping a stalled h2 connection is the right direction. The implementation needs changes before it can land:

1. **Hardcoded constants.** `30s / 15s / 60s` baked into the code is the main blocker. Every other transport knob here is configurable (`http_max_idle_conns*`, `http_write_buffer_size`, `http_idle_conn_timeout`), these must be too — `s3.http2_send_ping_timeout` / `http2_ping_timeout` / `http2_write_byte_timeout`, empty = disabled.

2. **Unconditional custom transport is a regression of #1376.** That guard was not cosmetic. `http.DefaultTransport.Clone()` gives Go's defaults — `MaxIdleConnsPerHost = 2`. The AWS SDK's own transport (`aws/transport/http/client.go`) uses `MaxIdleConnsPerHost = 10`, `MaxConnsPerHost = 2048`, plus its dialer timeouts and dial tracing. Since all `http_max_*` default to `0`, this PR silently drops every user — including plain AWS S3 — from 10 idle conns per host to 2, i.e. TCP+TLS re-established for most requests at `upload_concurrency > 2`. Build the transport from `awshttp.NewBuildableClient().GetTransport()` instead of `http.DefaultTransport`.

3. **`golang.org/x/net/http2` isn't needed.** This repo is `go 1.26`; `http.Transport.HTTP2` (Go 1.24+) exposes `SendPingTimeout` / `PingTimeout` / `WriteByteTimeout` for the bundled h2 stack, no extra dependency and no `ConfigureTransports` ordering concerns.

4. **`h2Err` is silently swallowed** — if `ConfigureTransports` fails the fix quietly does nothing.

5. For the record, scope: AWS S3 does not negotiate h2 over ALPN, so these settings only take effect on S3-compatible endpoints (Scaleway, MinIO, Ceph). Plain HTTP/1.1 stalls are still uncovered — no `ResponseHeaderTimeout` anywhere. Worth a follow-up, not a blocker here.

I'm landing an adjusted version of this with the config options and the SDK-default transport; will credit this PR. Thanks for the thorough goroutine dump — that's what made it diagnosable.

Slach (approved):

